Issue #26, opened 2022-10-17
Use connection handle in create_catalog and drop_catalog methods


ravjotbrar:
At the moment we call the login api directly. Need to use established connection and its parameters instead.

PR #29, opened 2022-10-17
Use connection handle in create_catalog and drop_catalog

ravjotbrar:
### Summary

Use connection handle in create_catalog and drop_catalog

### Description

Changed from directly calling the login api to using the connection we've already created. 

### Related Issue
https://github.com/dremio/dbt-dremio/issues/26

### Additional Reviewers
@jlarue26 
@ArgusLi


dbt/adapters/dremio/connections.py
```diff
@@ -258,12 +258,10 @@ def execute(
...
 
-        connection = self.get_thread_connection()
+        connection_param = self.get_thread_connection()
+        connection = self.open(connection_param)
         credentials = connection.credentials
-        api_parameters = self.build_api_parameters(credentials)
-
-        token = login(api_parameters)
-        connection.credentials.token = token
+        api_parameters = connection.handle._parameters
```
ArgusLi, line 264: Can we keep things consistent? Either have connection_parameters and api_parameters or connection_param and api_param.

jlarue26: What type does `self.get_thread_connection()` return? 
And let's use the full word if it is "parameters".

jlarue26: In Python, anything starting with an underscore is considered private. So we shouldn't call `connection.handle.**_**parameters`. 

I believe i created that private member variable. I don't think there is any reason to make it private, but let's think about if there would be any consequences of doing so (unlikely).

If we decide we are ok with sharing that info, then add a getter method.

dbt/adapters/dremio/connections.py
```diff
@@ -281,12 +279,10 @@ def drop_catalog(self, database, schema):
...
     def create_catalog(self, database, schema):
-        connection = self.get_thread_connection()
+        connection_param = self.get_thread_connection()
+        connection = self.open(connection_param)
         credentials = connection.credentials
-        api_parameters = self.build_api_parameters(credentials)
-
-        token = login(api_parameters)
-        connection.credentials.token = token
+        api_parameters = connection.handle._parameters
```
jlarue26, line 285: Same comments as for drop_catalog.

dbt/adapters/dremio/connections.py
ravjotbrar: How about just "connection"?

dbt/adapters/dremio/connections.py
```diff
@@ -258,10 +258,10 @@ def execute(
     def drop_catalog(self, database, schema):
         logger.debug('Dropping schema "{}.{}"', database, schema)
 
-        connection_param = self.get_thread_connection()
-        connection = self.open(connection_param)
+        connection = self.get_thread_connection()
```
ArgusLi, line 261: I don't think we should have connection parameters initialise with the same name as the thread connection. I see that both get_thread_connection() and open() return a Connection type (https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/base/connections.py), but we should still distinguish between the two. Maybe change the first one to thread_connection?

ravjotbrar: done

jlarue26 (commented):
Looks good, agree with how you resolved my comments.

jlarue26 (approved):

ArgusLi (approved):

Issue #22, opened 2022-10-13
Running all tests at once results in a ConnectionError


ravjotbrar:
Error message:

```requests.exceptions.ConnectionError: HTTPConnectionPool(host='172.25.2.87', port=9047): Max retries exceeded with url: /apiv2/login (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x0000018AB724F280>: Failed to establish a new connection: [WinError 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted'))```

Possible solution: https://stackoverflow.com/questions/23013220/max-retries-exceeded-with-url-in-requests

ravjotbrar:
### Verification
After setting-up tests/.env file with the correct environment variables,
- `python -m pytest tests/functional/adapter/test_basic.py -> 9/10 pass, 1 skipped (expected)`

Smoke Test:

- Ran dbt init sw using profile [2] -> pass
- dbt debug -> pass
- dbt run -> pass

### Required to run tests:

Install:
- `pip install pytest`
- `pip install python-dotenv`
- `pip install dbt-tests-adapter`

Create dbt-dremio/tests/.env file and add the following:

- `DREMIO_HOST=[Dremio Software cluster]`
- `DREMIO_USERNAME=[username]`
- `DREMIO_PASSWORD=[password]`
- `DREMIO_DATALAKE=[datalake source]`

### Future Suggestions:

@ArgusLi: I think a more elegant change to make would be to catch the exception being created and then doing retries with exponential backoff.

PR #31, opened 2022-10-18
Resolve hitting endpoint too many times

ravjotbrar:
### Summary

When running the tests consecutively, the login endpoint is hit too many times, resulting in a connection error. To resolve this we've added a delay between the tests.

### Description

Added a fixture that's autoused by the tests to sleep 1 second in between each tests.

### Related Issue

https://github.com/dremio/dbt-dremio/issues/22

### Additional Reviewers
@jlarue26 
@argusli

### DIFF
--- a/tests/functional/adapter/test_basic.py
+++ b/tests/functional/adapter/test_basic.py
@@ -1,4 +1,5 @@
 import pytest
+import time
 from tests.functional.adapter.utils.test_utils import (
     relation_from_name,
     check_relations_equal,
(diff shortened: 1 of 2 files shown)


ArgusLi (approved):
I think a more elegant change to make would be to catch the exception being created and then doing retries with exponential backoff. However, this also works and shouldn't add too much of a delay when running a small set of tests.

tests/functional/adapter/test_basic.py
```diff
@@ -40,6 +41,12 @@
 """
 
 
+@pytest.fixture(autouse=True)
+def slow_down_tests():
```
jlarue26, line 45: We should have a comment indicating why this is the case.
And perhaps we choose a different name: e.g. `throttle_login_connections` or something that also indicates why this is being done.

ravjotbrar: Fixed.

Issue #35, opened 2022-10-19
Create script to automate manual user testing


ravjotbrar:
Currently, we are running manual testing as a smoke test before creating PRs. To save time we need to create a script and integrate it as a workflow in Github actions.

PR #37, opened 2022-10-20
Add script for manual tests

ravjotbrar:
### Summary

Created a script to automate the manual smoke test we were doing before merging PRs. Next step is to call this using a github workflow so that we can automate testing as soon as a PR is created. 

### Description

Created a script which takes profiles.yml options as parameters. Another option is specified to decide if we're testing against software or cloud. These parameters will be fixed in our workflow, so no need to memorize the order of them to run the script manually. 

### Related Issue

https://github.com/dremio/dbt-dremio/issues/35

### Additional Reviewers

@ArgusLi 
@jlarue26

### DIFF
--- a/.github/scripts/smoke_test.sh
+++ b/.github/scripts/smoke_test.sh
@@ -1,0 +1,78 @@
+#! /bin/bash
+
+# Software test
+test_software() {
+cd ../../.. && \
+if [ -d "./$project" ]; then
+    rm -r $project
+fi
+dbt init -s $project && \
+{ cat > ./.dbt/profiles.yml << EOF
+$project:
+  outputs:
+    dev:
+      password: $test_password
+      port: 9047
+      software_host: $test_host
+      threads: 1
+      type: dremio
+      use_ssl: false
+      user: $test_user
+  target: dev
+EOF
+} && \
+
+cd $project/ && \
+dbt debug && \
+dbt run
+
+}
+
+#Cloud test
+test_cloud() {
+cd ../../.. && \
+if [ -d "./$project" ]; then
+    rm -r $project
+fi
+dbt init -s $project && \
+{ cat > ./.dbt/profiles.yml << EOF
+$project:
+  outputs:
+    dev:
+      cloud_host: $test_host
+      cloud_project_id: $test_cloud_project
+      pat: $test_token
+      threads: 1
+      type: dremio
+      use_ssl: true
+      user: $test_user
+  target: dev
+EOF
+} && \
+echo 'vars:' >> ./$project/dbt_project.yml && \
+echo '  dremio:reflections_enabled: false' >> ./$project/dbt_project.yml && \
+
+cd $project/ && \
+dbt debug && \
+dbt run 
+}
+
+# Main
+test_type=$1
+
+if [ $test_type == software ]; then 
+    test_user=$2
+    test_password=$3
+    test_host=$4
+    test_port=$5
+    project=software_proj
+    test_software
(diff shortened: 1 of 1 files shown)


.github/scripts/smoke_test.sh
```diff
@@ -0,0 +1,78 @@
...
+
+cd $project/ && \
+dbt debug && \
+dbt run 
+}
+
+# Main
+test_type=$1
+
+if [ $test_type == software ]; then
```
ArgusLi, line 63: We're missing testing PAT for software.

.github/scripts/smoke_test.sh
```diff
@@ -0,0 +1,78 @@
...
+      cloud_project_id: $test_cloud_project
+      pat: $test_token
+      threads: 1
+      type: dremio
+      use_ssl: true
+      user: $test_user
+  target: dev
+EOF
+} && \
+echo 'vars:' >> ./$project/dbt_project.yml && \
```
ArgusLi, line 52: Should we add this variable in software as well? Both are affected when testing with a non-admin account.

.github/scripts/smoke_test.sh
```diff
@@ -0,0 +1,78 @@
...
+{ cat > ./.dbt/profiles.yml << EOF
+$project:
+  outputs:
+    dev:
+      password: $test_password
+      port: 9047
+      software_host: $test_host
+      threads: 1
+      type: dremio
+      use_ssl: false
```
ArgusLi, line 19: Should we enable the option to have use_ssl be true?

ravjotbrar: Depends on the cluster we plan to use. At the moment we're using one that doesn't have ssl enabled.

.github/scripts/smoke_test.sh
ravjotbrar: I purposely left it out, since I thought one test for software and one for cloud were adequate for smoke tests. BUT I'm not against adding PAT for software.

.github/scripts/smoke_test.sh
ravjotbrar: Good point. I'll add it in.

.github/scripts/smoke_test.sh
ArgusLi: I think it'd be nice to have 3 test types: softwareUP, softwarePAT and cloud. I feel as though PAT is something we should test against as well.

.github/scripts/smoke_test.sh
ArgusLi: Although the current cluster we are testing against doesn't have ssl enabled, ssl enabled clusters will definitely be tested against in the future, and I feel as though having an extra variable would be nice here.

ArgusLi (approved):

jlarue26:
General comment: can we utilize more indenting for readability?

Here is a good guide, although I don't expect conformance: https://lug.fh-swf.de/vim/vim-bash/StyleGuideShell.en.pdf

ravjotbrar:
> General comment: can we utilize more indenting for readability?
> 
> Here is a good guide, although I don't expect conformance: https://lug.fh-swf.de/vim/vim-bash/StyleGuideShell.en.pdf

I tried adding indenting, but it breaks the script for some reason.

.github/scripts/smoke_test.sh
```diff
@@ -0,0 +1,124 @@
+#! /bin/bash
+
```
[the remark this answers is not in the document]
ravjotbrar: Done.

Issue #18, opened 2022-10-11
HIGH PRIORITY (workaround): Automatically update `dbt_project.yml` to disable reflections when connecting with non-admin account


jlarue26:
When using a non-admin account, the following needs to be added to the bottom of the `dbt_project.yml` file at the root of the Dremio Cloud dbt project directory:

```
vars:
  dremio:reflections_enabled: false
```

This should be added automatically.

ArgusLi:
Also figure out why it is doing sys table reflection calls.

ArgusLi:
@fabrice-etanchaud thanks for the information you provided. We are concerned with most users of the adapter not having admin privileges. It appears that there is no way to access reflections without admin privileges. Therefore the question is how to by default disable the reflection materialisation instead of the current default which enables it. 

Would you happen to know?

fabrice-etanchaud:
Easier said than done : If there is no such restriction with the API, now that you added API support, the call to the reflections system table in the catalog macros could be replaced with an API call to get the list of reflections, and the results unioned with the SQL based VDS and table list. Or maybe get the whole list of reflections + tables + vds via API ?

fabrice-etanchaud:
Maybe we could replace true by false here : 

https://github.com/dremio/dbt-dremio/blob/340bfcefa282bd575d202cc8b2d07c122813c2f5/dbt/include/dremio/macros/adapters/metadata.sql#L28

and here : 

https://github.com/dremio/dbt-dremio/blob/340bfcefa282bd575d202cc8b2d07c122813c2f5/dbt/include/dremio/macros/materializations/reflection/reflection.sql#L16

I hope I can find more time to help next week, I will be on vacation.

ArgusLi:
Even looking at the Rest API documentation, listing all tables is admin only, however users without admin privileges can look at reflection summaries.

Thanks for the information Fabrice, that saves me a lot of time. I think having reflections defaulting to off is only a temporary solutions, we'd definitely like to provide reflections. Just using drop and create is definitely an interesting approach. I'll take a closer look at it, thank you!

Enjoy your vacation!

PR #38, opened 2022-10-21
Make dremio:reflections_enabled default to false if not declared.

ArgusLi:
### Summary

Make dremio:reflections_enabled default to false if not declared.

### Description

dremio:reflections_enabled will now default to false. We expect most users to have accounts without admin privileges. Since listing reflections requires admin privileges, the default should be to disable reflections, with admins having to enable them if needed.

### Related Issue

#18 

### Additional Reviewers

@jlarue26 @ravjotbrar

### DIFF
--- a/dbt/include/dremio/macros/adapters/metadata.sql
+++ b/dbt/include/dremio/macros/adapters/metadata.sql
@@ -25,7 +25,7 @@ limitations under the License.*/
 
     with cte as (
 
-    {%- if var('dremio:reflections_enabled', false) %}
+    {%- if var('dremio:reflections_enabled', default=false) %}
 
       select
         case when position('.' in table_schema) > 0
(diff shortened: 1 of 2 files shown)


dbt/include/dremio/macros/adapters/metadata.sql
```diff
@@ -140,7 +140,7 @@ limitations under the License.*/
         + (('.' + schema) if schema != 'no_schema' else '') -%}
   {% call statement('list_relations_without_caching', fetch_result=True) -%}
 
-    {%- if var('dremio:reflections_enabled', true) -%}
+    {%- if var('dremio:reflections_enabled', false) -%}
```
ravjotbrar, line 143: The logic here seems like it'll select from sys.reflections if dremio:reflections_enabled is set to false?

ArgusLi: That's what I thought initially as well, the second argument in var() is actually `default`, which is the default value for the variable if one is not explicitly defined.
https://docs.getdbt.com/reference/dbt-jinja-functions/var#variable-default-values

ravjotbrar: Can we use a keyword argument, i.e default=false?

ArgusLi: Added in commit.

ravjotbrar (approved):

PR #47, opened 2022-10-24
Get functional tests working against Dremio Cloud

ravjotbrar:
### Summary

Made changes to the tests to refer to the S3 bucket instead of the GCS bucket. The source name is still the same, but the root_path in the tests will now prefix the bucket name, instead of the source name. In addition, I've added a dremio cloud profile and dremio software profile to run the functional tests against. Since the privileges table is different in Dremio Software and Dremio Cloud, the grants macro needs to check which profile we're using in order to get data from the privileges table. 

#### Grants Macro
- Check to see if we're using Dremio Cloud or Dremio Software as our host and set the privileges_table name accordingly
- Quote the grantee user, as Dremio Cloud users have dots in their name

#### Functional Tests Configuration
- Add a new parser option for pytest called `--profile`
- Add two profile targets: dremio_cloud_target() and dremio_software_target()

These changes allow us to specify 'dremio_cloud' or 'dremio_software' as the profile when running the functional tests

#### Datalake Constant
- Rename to 'BUCKET' and add another constant called 'SOURCE' to differentiate between the two in Dremio

### Additional Reviewers

@jlarue26 
@ArgusLi


ravjotbrar:
workflow

dbt/include/dremio/macros/adapters/apply_grants.sql
```diff
@@ -22,18 +22,25 @@ limitations under the License.*/
...
         {%- set relation_without_double_quotes = relation.database ~ '.' ~ relation.schema ~ '.' ~ relation.identifier-%}
     {%- endif %}
-
+    
+    {%- if target.cloud_host and not target.software_host -%}
+        {%- set privileges_table = 'sys.project.privileges' -%}
+    {%- elif target.software_host and not target.cloud_host -%}
+        {%- set privileges_table = 'sys.privileges' -%}
+    {%- else -%}
+        {%- set privileges_table = 'sys.privileges' -%}
```
ArgusLi, line 31: Can we throw an error here?

ArgusLi (commented):
Everything else looks amazing. I am very excited for these changes!

dbt/include/dremio/macros/adapters/apply_grants.sql
ravjotbrar: Good suggestion. Done.

ArgusLi (approved):

Issue #34, opened 2022-10-18
Modify URL Builder to use less parameters


ArgusLi:
Too many parameters and passing stuff that is optional.
Should pass object that it knows is cloud.

ArgusLi:
Final Design:
![image](https://user-images.githubusercontent.com/43020525/198379458-5a7df0ee-cb5c-4716-b4be-f8136de35847.png)

ArgusLi:
When implemented, noticed a deficiency in build(). New design: 
![newUrlBuilder drawio (3)](https://user-images.githubusercontent.com/43020525/198411595-f291b432-95de-47fa-85f8-78752910a62d.png)

PR #50, opened 2022-10-27
Implement new url builder.

ArgusLi:
### Summary

Implement new url builder.

### Description

- Create dbt/adapters/dremio/api/parameters and move authentication.parameters into said parameters module.
- Create ParametersBuilder and its subclasses CloudParametersBuilder and SoftwareParametersBuilder.
- Modify Parameters and create two new subclasses CloudParameters and SoftwareParameters.
- Replace DremioConnectionManager.build_api_parameters() with ParametersBuilder.getParameters().
- Move DremioCredentials into its own module to resolve circular dependencies.
- Update imports and uses.

Had to make changes to the finalised class diagram to accept credentials instead of multiple variables.
New diagram:
newUrlBuilder drawio (3)

### Related Issue

https://github.com/dremio/dbt-dremio/issues/34


dbt/adapters/dremio/api/rest/endpoints.py
```diff
@@ -205,28 +189,20 @@ def get_catalog_item(
...
 def delete_catalog(api_parameters, cid, ssl_verify=True):
 
-    url = UrlBuilder.catalog_url(
-        api_parameters.base_url,
-        api_parameters.is_cloud,
-        api_parameters.cloud_project_id,
-    )
+    url = UrlBuilder.catalog_url(api_parameters)
     return _delete(
         url + f"/{cid}",
```
jlarue26, line 207: `cid` should be passed to UrlBuilder to take care of.

dbt/adapters/dremio/api/rest/url_builder.py
```diff
@@ -31,92 +34,90 @@ class UrlBuilder:
     CLOUD_CATALOG_ENDPOINT = CLOUD_PROJECT_ENDPOINT + "/{}/catalog"
 
     @classmethod
-    def login_url(cls, base_url):
-        return base_url + UrlBuilder.SOFTWARE_LOGIN_ENDPOINT
+    def login_url(cls, parameters: Parameters):
+        return parameters.base_url + UrlBuilder.SOFTWARE_LOGIN_ENDPOINT
```
jlarue26, line 38: Should we raise an error or warn if we get passed a CloudParameters object? If anything, that would provide documentation as to why there is no cloud login endpoint.

dbt/adapters/dremio/api/rest/url_builder.py
```diff
@@ -31,92 +34,90 @@ class UrlBuilder:
...
             )
-        return base_url + UrlBuilder.SOFTWARE_JOB_ENDPOINT + "/" + job_id
+        return parameters.base_url + UrlBuilder.SOFTWARE_JOB_ENDPOINT + "/" + job_id
 
     @classmethod
-    def job_cancel_url(cls, base_url, job_id, is_cloud=False, offset=0, limit=100):
-        url_path = None
-        if is_cloud:
-            url_path = base_url + UrlBuilder.CLOUD_JOB_ENDPOINT
+    def job_cancel_url(cls, parameters: Parameters, job_id, offset=0, limit=100):
```
jlarue26, line 60: Remove offset and limit as they are not appropriate for this call (and are never used).
(I understand these were pre-existing but we should get rid of them.)

dbt/adapters/dremio/api/rest/url_builder.py
```diff
@@ -31,92 +34,90 @@ class UrlBuilder:
...
 
     @classmethod
     def job_results_url(
         cls,
-        base_url,
+        parameters: Parameters,
         job_id,
-        is_cloud=False,
         offset=0,
         limit=100,
```
jlarue26, line 75: Technically this is a magic number. Put I am ok if we leave as is in this particular usage.

dbt/adapters/dremio/api/parameters.py
```diff
@@ -0,0 +1,112 @@
...
+        if credentials.cloud_host != None and credentials.cloud_project_id != None:
+            return CloudParametersBuilder(
+                cls.__buildDremioAuthentication(credentials=credentials),
+                credentials.cloud_host,
+                credentials.cloud_project_id,
+            )
+        elif (
+            credentials.software_host != None
+            and credentials.port != None
+            and credentials.use_ssl != None
```
ravjotbrar, line 59: Isn't use_ssl populated by default for both cloud and software?

ravjotbrar: Same with port

ArgusLi: I think checking the three parameters to be used is good defensive programming. Given that all of the variables are absolutely necessary in the UrlBuilder, and that users can change profiles.yml, I think it's good practice to ensure that they are all set.

Port also does not always default to 9047.[ Server admins can change ports in dremio.conf](https://docs.dremio.com/software/advanced-administration/dremio-conf/).

dbt/adapters/dremio/api/rest/url_builder.py
ArgusLi: Could the same goal be achieved by changing the method signature to `def login_url(cls, software_parameters: SoftwareParameters):`?

dbt/adapters/dremio/api/rest/url_builder.py
ArgusLi: Would adding class constants and leaving a comment help? e.g. 
``` 
#https://docs.dremio.com/software/rest-api/jobs/get-job/
OFFSET_DEFAULT = 0
LIMIT_DEFAULT = 100
```
 Then replacing the default magic numbers with:
 ```
 def job_results_url(
        cls,
        parameters: Parameters,
        job_id,
        offset=OFFSET_DEFAULT,
        limit=LIMIT_DEFAULT,
    ):
```

File: STYLE_GUIDE.md [hunk 6]
```diff
@@ -0,0 +1,10 @@
+# WIP
```
jlarue26, line 1: Make sure we add the copyright please :-)

dbt/adapters/dremio/api/handle.py
```diff
@@ -46,3 +41,8 @@ def close(self):
     def rollback(self):
         # todo
         logger.debug("Handle rollback not implemented.")
+
+    def __init__(self, parameters: Parameters):
```
jlarue26, line 45: This is a constructor should should be before public methods according to our style guide, no?

ArgusLi: Oh whoops, got a little too excited with moving methods. I'll fix it, thanks.

dbt/adapters/dremio/api/parameters.py
ravjotbrar: In our dbt adapter, port and use_ssl will always default to a value (true and 9047, respectively), which means they will never be None, unless we make that change in DremioCredentials specifically.

ravjotbrar: My point is that port will default to a value, doesn't matter if its 9047 or not

ArgusLi: Ah I see what you mean, it's defined in DremioCredentials and default to the values you've listed. I think I'd still prefer to keep the checks for legibility and just being defensive. If someone changes the DremioCredentials and no longer defaults them to a certain value, having this check would prevent a problem further down and lead to easier troubleshooting.

ravjotbrar: We still need a case that checks for if someone provides a software_host and cloud_host. In this case, the software checks will pass and build software params.

ArgusLi: You make a good point. Let me make some changes.

dbt/adapters/dremio/api/rest/url_builder.py
ArgusLi: Add comment why we're only accepting SoftwareParameters.

ravjotbrar (approved):

Issue #43, opened 2022-10-24
dbt-core 1.2 upgrade: BaseDocsGenerate and BaseDocsGenReferences


jlarue26:
From https://github.com/dbt-labs/dbt-core/discussions/5468:

> https://github.com/dbt-labs/dbt-core/pull/5058 is another step along https://github.com/dbt-labs/dbt-core/issues/4788 to the new framework in order to empower adapter maintainers and other contributors to make use of the same tests that the core team uses for their own adapters. Effectively, this test is validates an adapter's ability to correctly generate the catalog that serves as the static backend of a project docs site.
> If your adapter does not add extra relation-level metadata (e.g. table size (rows + bytes), last modified timestamp) which is the case by default, then you can follow the same inherit and pass pattern to enable your version of BaseDocsGenerate and BaseDocsGenReferences. However, if you are supplementing the catalog with more metadata, you'll have to:
> 
> Add a method that defines stats for this adapter [e.g. dbt-bigquerys](https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/expected_stats.py)
> Reimplement the expected_catalog fixture, [passing the above into model_stats and seed_stats](https://github.com/dbt-labs/dbt-bigquery/blob/0212fd621ede4c24929a008de718a7e45bc32cec/tests/functional/adapter/test_basic.py#L68-L81)
> Example PRs:
> 
> https://github.com/dbt-labs/dbt-bigquery/pull/190
> https://github.com/dbt-labs/dbt-redshift/pull/116

ravjotbrar:
Since, default implementations are automatically inherited by adapters and included in the testing suite, any issues related to these macros would become apparent when running the tests. All tests  currently pass and any issues (including timestamp) were addressed in [basic func tests](https://dremio.atlassian.net/browse/DX-56929). Closing.

Issue #36, opened 2022-10-19
Remove snapshot staging macro


ravjotbrar:
Remove dremio__snapshot_staging_table, as this was added in to view some logs. It's the same implementation dbt-core provides as the default, therefore it provides no value.

PR #52, opened 2022-10-28
Add BaseDocsGenerate and BaseDocsGenReferences Tests

ravjotbrar:
### Summary

Added these tests as part of the requirements for upgrading to dbt-core 1.2. 

#### BaseDocsGenerate
- Modify second_model.sql to explicitly use alternate schema
- Modify models fixture to use our version of second_model.sql
- Modify verify_catalog from dbt-core to exclude source table verification
- Modify unique_schema fixture to prepend schema with rav-test folder
- Modify project_config_update fixture to prevent the seed being created as a view
- Modify profiles fixture to make root_path the same as schema
- Modify expected_catalog fixture to accommodate dremio data types
- Include test_run_and_generate methods so they use our version of verify_catalog

#### BaseDocsGenReferences
- Modify project_config_update fixture to allow creating a view for seeds. This is because the ephemeral_summary models looks for the seed under database.schema (which is only used for views). 
- Modify unique_schema, profiles, and expected catalog fixtures for same reasons as above

#### Utilities
- Modify base_expected_catalog to look for datalake instead of database 
- Modify expected_references_catalog for the same reason above

#### Other
- Relevant imports
- Remove staging snapshot macro

### Related Issue

https://github.com/dremio/dbt-dremio/issues/43
https://github.com/dremio/dbt-dremio/issues/36

### Additional Reviewers

@jlarue26 
@ArgusLi


tests/functional/adapter/test_basic.py
```diff
@@ -495,3 +512,200 @@ def dbt_profile_data(
...
+            "model.sql": models__model_sql,
+        }
+
+    # Override this fixture to prepend our schema with rav-test
+    # This ensures the schema works with our datalake
+    @pytest.fixture(scope="class")
+    def unique_schema(self, request, prefix) -> str:
+        test_file = request.module.__name__
+        test_file = test_file.split(".")[-1]
+        unique_schema = f"rav-test.{prefix}_{test_file}"
```
ravjotbrar, line 572: rav-test should be constant

tests/functional/adapter/test_basic.py
```diff
@@ -495,3 +512,200 @@ def dbt_profile_data(
...
+        alternate_schema = unique_schema + "_test"
+        return {
+            "asset-paths": ["assets", "invalid-asset-paths"],
+            "vars": {
+                "test_schema": unique_schema,
+                "alternate_schema": alternate_schema,
+            },
+            "seeds": {
+                "quote_columns": True,
+                "+twin_strategy": "clone",
```
ravjotbrar, line 668: Could remove as clone is default

tests/functional/adapter/test_basic.py
```diff
@@ -495,3 +512,200 @@ def dbt_profile_data(
...
+            },
+            "models": {
+                "+twin_strategy": "prevent",
```
ravjotbrar, line 671: Why is this necessary?

ravjotbrar (commented):
Try to avoid code duplication

tests/functional/adapter/utils/test_utils.py
```diff
@@ -104,3 +105,264 @@ def check_relations_equal(adapter, relation_names: List, compare_snapshot_cols=F
...
+    case=None,
+    case_columns=False,
+):
+
+    if case is None:
+
+        def case(x):
+            return x
+
+    col_case = case if case_columns else lambda x: x
```
ArgusLi, line 132: Sorry, what's the purpose of lambda x: x?

ravjotbrar: This was written by dbt, so I'm not entirely sure. Based on my limited interaction with this method, I assume it has to do with the case of specific column types.

tests/functional/adapter/test_basic.py
ravjotbrar: Done

tests/functional/adapter/test_basic.py
ravjotbrar: Removed.

tests/functional/adapter/basic/test_base_mat.py
```diff
@@ -14,13 +14,14 @@
...
+from tests.functional.adapter.utils.test_utils import DATALAKE
 
-
+# Unable to insert variable into docstring, so "rav-test" is hardcoded
 schema_base_yml = """
 version: 2
 sources:
   - name: raw
-    database: "rav-test"
+    database: "rav-test"
```
ArgusLi, line 24: should this not be DATALAKE as well?

ravjotbrar: I tried to insert a variable, but it turns out inserting variables into docstrings isn't trivial, especially when the "{" notation is also used to insert variables from the dbt config.

PR #97, opened 2022-12-08
Implement job results pagination

### LINKED ISSUE #61: Raw reflection limits display columns to 100
## Description
When creating a raw reflection on a view that has more than 100 columns **without** specifying the `display` configuration, the resulting raw reflection only has the first 100 display columns.

The following model is used to create the reflection:
```sql
{{
  config(
    materialized = 'reflection',
    schema = 'test',
    reflection_type = 'raw'
  )
}}

-- depends_on: {{ ref('view_with_over_100_columns') }}
```
The `dbt.log` shows that only the first 100 columns are used in the display section:
```sql
  alter dataset "db"."test"."view_with_over_100_columns"
    create raw reflection "ref_raw_view_with_over_100_columns"
    using
      display ("<only the first 100 columns>")
```
Interestingly, the macro [`dremio__get_columns_in_relation`](https://github.com/dremio/dbt-dremio/blob/main/dbt/include/dremio/macros/adapters/columns.sql#L15) that is used to get the columns uses a query that returns all available columns:
```sql
select column_name as column_name
    ,lower(data_type) as data_type
    ,character_maximum_length
    ,numeric_precision
    ,numeric_scale
from information_schema.columns
where ilike(table_schema, 'db.test')
and ilike(table_name, 'view_with_over_100_columns')
order by ordinal_position
```

## Expectation
I would expect all available columns to be used as display columns when the `display` configuration is not specified in the model.

## Environment

**dbt-dremio version:** 1.1.0
**dbt-core version:** 1.3.0
**Python:** 3.8.5
**OS:** Linux 18a2b7f4c743 6.0.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 03 Nov 2022 18:01:58 +0000 x86_64 GNU/Linux

ArgusLi:
### Summary

Implement job results pagination

### Description

- Add optional row_limit argument to _populate_job_results().
- Set row_limit default to 100.
- Create test_job_result.py unit test that tests pagination.

#### smoke_test
softwareUP --> PASS
softwarePAT --> PASS

### Changelog

### Related Issue

Resolves #61

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
 
 -   Override dbt-core `default__type_string()` macro to use Dremio Supported VARCHAR instead of the default string. ([#80](https://github.com/dremio/dbt-dremio/pull/80))
 
--   Change \_populate_job_results() to have an optional row_limit argument with default set to 100 (Dremio's default).
+-   Change \_populate_job_results() to have an optional row_limit argument with default set to 100 (Dremio's default). ([#61](https://github.com/dremio/dbt-dremio/issues/61))
 
--   Implement pagination in \_populate_job_results()
+-   Implement pagination in \_populate_job_results() ([#61](https://github.com/dremio/dbt-dremio/issues/61))
 
 ## Under the Hood
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 -   Override dbt-core `default__type_string()` macro to use Dremio Supported VARCHAR instead of the default string. ([#80](https://github.com/dremio/dbt-dremio/pull/80))
 
--   Change \_populate_job_results() to have limit of 500 (Dremio's limit).
+-   Change \_populate_job_results() to have page_limit of 100 (Dremio's default).
+
+-   Change \_populate_job_results() to have an additional optional page_limit argument.
 
 ## Under the Hood
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
 
 -   Override dbt-core `default__type_string()` macro to use Dremio Supported VARCHAR instead of the default string. ([#80](https://github.com/dremio/dbt-dremio/pull/80))
 
--   Change \_populate_job_results() to have page_limit of 100 (Dremio's default).
+-   Change \_populate_job_results() to have an optional page_limit argument with default set to 100 (Dremio's default).
 
--   Change \_populate_job_results() to have an additional optional page_limit argument.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 -   Override dbt-core `default__type_string()` macro to use Dremio Supported VARCHAR instead of the default string. ([#80](https://github.com/dremio/dbt-dremio/pull/80))
 
--   Change \_populate_job_results() to have limit of 500 (Dremio's limit).
+-   Change \_populate_job_results() to have an optional row_limit argument with default set to 100 (Dremio's default).
+
+-   Implement pagination in \_populate_job_results()
```
ravjotbrar, line 11: Let's link the relevant issue here

dbt/adapters/dremio/api/cursor.py
```diff
@@ -147,11 +147,31 @@ def _populate_rowcount(self):
...
-            self._job_results = job_results(
-                self._parameters, self._job_id, offset=0, limit=500, ssl_verify=True
+            combined_job_results = job_results(
+                self._parameters,
+                self._job_id,
+                offset=0,
+                limit=row_limit,
+                ssl_verify=True,
             )
+            total_row_count = combined_job_results["rowCount"]
```
ravjotbrar, line 159: Is there ever a case where "rowCount" isn't available for a job?

tests/unit/test_job_results.py
```diff
@@ -0,0 +1,78 @@
...
+            {"id": 1},
+            {"id": 2},
+            {"id": 3},
+            {"id": 4},
+            {"id": 5},
+            {"id": 6},
+            {"id": 7},
+        ],
+    }
+
```
ravjotbrar, line 62: Is there a reason we're not adding a test for the greater than 100 cols issue?

ArgusLi: As we discussed, I think that having a unit test that ensures pagination works correctly tests the root cause of the issue. Therefore, there is no need to have a dedicated tests for >100 columns.

CHANGELOG.md
ArgusLi: Good catch.

dbt/adapters/dremio/api/cursor.py
ArgusLi: _populate_job_results() will only execute after _populate_rowcount(). In _populate_rowcount() if rowCount is not in the job_status_response, it means that no rows will be returned, and rowCount will be set to -1. 

I have manually verified that this does not affect pagination.

Issue #44, opened 2022-10-24
dbt-core 1.3 upgrade: Incremental mats: more standard and more error-proof


jlarue26:
From https://github.com/dbt-labs/dbt-core/discussions/6011: 

> user-facing changes
> new in 1.3:
> breaking change. the default strategy of incremental models used to be delete+insert. it is now append. See point below to override this.
> a new macro, adapter_get_incremental_default_sql(), in which adapter maintainers can specify a 'default' strategy for incremental models.
> a new Adapter method, valid_incremental_strategies() in which adapter maintainers can define all supported strategies. Previously, if i user defined an unsupported incremental strategy, they would not learn of their error until compile time or later. Now dbt will throw an error before attempting to compile.
> non-user facing changes
> there was significant work (https://github.com/dbt-labs/dbt-core/issues/5245 https://github.com/dbt-labs/dbt-core/pull/5359) put into refactoring the default incremental materialization. If your adapter overrides the default, we recommend that you either:
> 2. refactor your materialization to match that of the default
> even better, determine exactly why you currently need to override the default, and perhaps we can find a way to bring that flexibility into the core adapter so that you can stay up to date with the default mat

ArgusLi:
Not in the scope of our current dbt strategy.

PR #109, opened 2023-01-25
44 dbt core 13 upgrade incremental mats more standard and more error proof

ArgusLi:
### Summary

Upgrade incremental materialization to new dbt-core 1.3 logic while retaining Dremio specific changes.

### Description

- Add valid_incremental_strategies() with append being the only valid strategy.
- Refactor materialization to match new default incremental materialization.
- Add macro dremio__rename_relation() that uses CTAS and DROP instead of default ALTER TABLE and RENAME to.

### Test Results

30/30 tests passed.

### Changelog

### Related Issue
[44](https://github.com/dremio/dbt-dremio/issues/44)


dbt/include/dremio/macros/materializations/incremental/incremental.sql
```diff
@@ -14,64 +14,96 @@ limitations under the License.*/
...
+    
   {% endif %}
 
   {%- call statement('main') -%}
     {{ build_sql }}
   {%- endcall -%}
 
-  {% if not(old_relation is none or full_refresh_mode) %}
-    {{ adapter.drop_relation(target_tmp_relation) }}
+  {% if need_swap %}
```
ravjotbrar, line 78: I noticed rename_relation is being used in view and table macros as well. Is this something we should be doing as well?

ravjotbrar: This might be a separate ticket, but it'll keep our macros more up to date.

dbt/include/dremio/macros/materializations/incremental/incremental.sql
```diff
@@ -14,64 +14,96 @@ limitations under the License.*/
...
+    {% set build_sql = get_create_table_as_sql(False, target_relation, sql) %}
   {% elif full_refresh_mode %}
-    {% do adapter.drop_relation(old_relation) %}
-    {% set build_sql = create_table_as(False, target_relation, external_query(sql)) %}
+    {% set build_sql = get_create_table_as_sql(False, intermediate_relation, sql) %}
+    {% set need_swap = true %}
   {% else %}
-    {% if tmp_relation is not none %}
-      {{ adapter.drop_relation(tmp_relation) }}
+    {% do run_query(get_create_table_as_sql(True, temp_relation, sql)) %}
```
ravjotbrar, line 54: the old run_query had the sql wrapped in an external_query macro. Is there a reason we removed this?

dbt/include/dremio/macros/materializations/incremental/incremental.sql
```diff
@@ -14,64 +14,96 @@ limitations under the License.*/
...
+  {{ run_hooks(pre_hooks, inside_transaction=False) }}
+ 
+  -- `BEGIN` happens here:
+  {{ run_hooks(pre_hooks, inside_transaction=True) }}
 
-  {{ run_hooks(pre_hooks) }}
+  {% set to_drop = [] %}
 
-  {% if old_relation is none %}
-    {% set build_sql = create_table_as(False, target_relation, external_query(sql)) %}
```
ravjotbrar, line 45: The old build_sql  had the sql wrapped in an external_query macro. Is there a reason we removed this?

dbt/include/dremio/macros/materializations/incremental/incremental.sql
```diff
@@ -14,64 +14,96 @@ limitations under the License.*/
...
 {% materialization incremental, adapter='dremio' -%}
 
-  {%- set identifier = model['alias'] -%}
-  {%- set old_relation = adapter.get_relation(database=database, schema=schema, identifier=identifier) -%}
-  {%- set target_relation = api.Relation.create(identifier=identifier,
-                                                schema=schema,
-                                                database=database,
-                                                type='table') -%}
-
-  {%- set tmp_identifier = model['name'] + '__dbt_tmp' -%}
```
ravjotbrar, line 24: Is there a reason we're getting rid of the tmp_identifier?

dbt/include/dremio/macros/materializations/incremental/incremental.sql
```diff
@@ -14,64 +14,96 @@ limitations under the License.*/
...
-                                                database=database,
-                                                type='table') -%}
-
-  {%- set raw_strategy = config.get('incremental_strategy', validator=validation.any[basestring]) or 'append' -%}
-  {%- set raw_file_format = config.get('format', validator=validation.any[basestring]) or 'iceberg' -%}
-  {%- set file_format = dbt_dremio_validate_get_file_format(raw_file_format) -%}
-  {%- set strategy = dbt_dremio_validate_get_incremental_strategy(raw_strategy, file_format) -%}
+  -- relations
+  {%- set existing_relation = load_cached_relation(this) -%} 
+  {%- set target_relation = this.incorporate(type='table') -%}
```
ravjotbrar, line 19: We should be careful about using dbt's defaults for these macros. In the end, we as a team need to add more dremio specific test cases around custom schemas and different targets, to ensure changes like these don't break any of the edge cases.

dbt/include/dremio/macros/materializations/incremental/incremental.sql
ArgusLi: rename_relation is used here for full_refresh_mode to fully replace the relation. I don't see a real need to use it in other macros unless its functionality is needed.

dbt/include/dremio/macros/materializations/incremental/incremental.sql
ArgusLi: tmp_identifier was used in the old logic to get tmp_relation and create target_tmp_relation. The new logic has moved to temp_relation and intermediate_relation with a different relation creation strategy, therefore, it is no longer needed.

dbt/include/dremio/macros/adapters/relation.sql
```diff
@@ -25,3 +25,12 @@ limitations under the License.*/
     drop {{ relation.type }} if exists {{ relation }}
   {%- endcall %}
 {% endmacro %}
+
+{% macro dremio__rename_relation(from_relation, to_relation) -%}
+  {% call statement('rename_relation1/2 - create to_relation from from_relation') -%}
+    CREATE TABLE {{ to_relation }} AS (select * from {{from_relation}})
+  {%- endcall %}
+  {% call statement('rename_relation2/2 - drop from_relation') -%}
```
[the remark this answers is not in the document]
ArgusLi: I didn't think of that. Good catch. However, I do not like the dremio__create_table_as macro, there is an extra parameter "temporary" that does not get used at all in the macro. I will give both a try.

dbt/include/dremio/macros/materializations/incremental/incremental.sql
ArgusLi: On the contrary, I think incorporating new dbt logic is necessary. The new dbt logic is more elegant and less "hacky". It's also not like we're just using dbt's default, we're using its logic but keeping Dremio specific changes. In the future I think the ideal direction is actually using dbt's defaults for everything and just having Dremio specific macros. That way we do not have to maintain any of these materialization strategies - that will be done by the dbt team.

However, I agree we should have more Dremio specific test cases.

dbt/include/dremio/macros/materializations/incremental/incremental.sql
ArgusLi: We really should be using external_query. I will change this. Thanks!

dbt/include/dremio/macros/materializations/incremental/incremental.sql
ArgusLi: We really should be using external_query. I will change this. Thanks!

dbt/include/dremio/macros/adapters/relation.sql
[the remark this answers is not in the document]
ArgusLi: dremio__create_table() has been implemented, but dremio__drop_table() leads to an error and will be addressed in https://github.com/dremio/dbt-dremio/issues/113.

ravjotbrar (approved):

Issue #91, opened 2022-11-24
Running all the functional tests at once causes TestModelGrants to fail


ravjotbrar:
The error is `Unable to create or update view. Filesystem views are not supported.`

This error is caused by the job attempting to create a view in the object storage source rather than a space for this specific test. 

Expected Query:
```
  create or replace view "**dbt_test_space**"."test16693270750636647031_test_model_grants"."my_model" as (
    select 1 as fun
  )
```

Actual Query:
```
  create or replace view "**dbt_test_source**"."test16693270750636647031_test_model_grants"."my_model" as (
    select 1 as fun
  )
```

NOTE: This error is not present when running the grants tests separate from all the other tests.

Issue #128, opened 2023-02-10
Create issue template


ravjotbrar:
### Description

Create issue templates that reporter's can use to write bug reports or enhancement requests.

### Justification

This ensures we get all the required information when deciding to triage and pursue an issue.

PR #129, opened 2023-02-10
Create issue templates 

ravjotbrar:
### Summary

Created issue templates for bug reports and enhancement requests. In this same PR, I also moved three individual tests to the basic tests folder, so that these are easier to run (until we fix https://github.com/dremio/dbt-dremio/issues/91) 

### Description

Bug reports now outline information from the reporter's perspective:
1. Current Behaviour
2. Expected Behaviour
3. Steps to reproduce
4. Adapter Version
5. Relevant logs

Enhancement requests are simpler with:
1. Description of the enhancement
2. Justification for the enhancement

### Test Results

N/A

### Changelog

N/A

### Related Issue

https://github.com/dremio/dbt-dremio/issues/128


ArgusLi:
I like that we're reorganising tests, but I think we should match the structure of dbt-core.

ravjotbrar:
> I like that we're reorganising tests, but I think we should match the structure of dbt-core.

I agree with that sentiment. At the moment it's just easier running the tests manually when they're in two separate folders. I suppose we could also write a script that runs them separately as well (only until we get the issue resolved where we can't run all the func tests at once).

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,55 @@
...
+      placeholder: |
+        1. In this environment...
+        2. With these relevant config options...
+        3. Run '...'
+        4. See error...
+  - type: dropdown
+    id: version
+    attributes:
+      label: Adapter Version
+      description: What version of our software are you running?
```
ArgusLi, line 39: Maybe "what version of dbt-dremio"?

ArgusLi: Also should we have one for dremio version as well?

ArgusLi (commented):
LGTM apart from that - just some nits

.github/ISSUE_TEMPLATE/bug_report.yml
ravjotbrar: Good suggestions. Will add them in.

jlarue26 (approved):
Looks great, thanks!

PR #147, opened 2023-03-03
134 performance enhancement for information schematables

ArgusLi:
### Summary

Performance Enhancement for information schematables

### Description

- Create a new variable `dremio:exact_search_enabled` which defaults to false
- If variable set to true, changes the ilike to a basic equality for the `dremio__list_relations_without_caching` macro when reflections are disabled (which is the default)

### Test Results

Thanks @ravjotbrar for running the test suite with the variable not set (which is the default).
All tests pass.

### Changelog


ravjotbrar (approved):
Assuming we're going to address/test all the other ilikes in another PR, I think everything here is good, especially since this isn't a breaking change.

PR #181, opened 2023-05-03
Update dev_requirements.txt and THIRD_PARTY_LICENSES.txt for sqlparse…

ArgusLi:
… v0.4.4.

### Summary

Upgrade sqlparse to v0.4.4

### Description

Upgrade sqlparse to v0.4.4 to fix security vulnerability

### Test Results

Smoke tests passed successfully.

### Changelog

-   [X] Added a summary of what this PR accomplishes to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,21 @@
+# dbt-dremio 1.X.X - release TBD
+
+## Features
+
+## Fixes
+
+## Under the Hood
+
+## Dependency
+
+-   Upgrade sqlparse to 0.4.4 [#180](https://github.com/dremio/dbt-dremio/issues/180).
+
 # dbt-dremio 1.4.5 - release March 23, 2023
 
 ## Features
 
 ## Fixes
+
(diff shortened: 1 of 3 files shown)


ravjotbrar (commented):
Do we need sqlparse as a requirement?

ArgusLi:
> Do we need sqlparse as a requirement? I wonder how many of these packages we could remove.

dbt-core requires sqlparse - this can be seen [here](https://github.com/dbt-labs/dbt-core/blob/main/core/setup.py).

ravjotbrar:
> dbt-core requires sqlparse - this can be seen [here](https://github.com/dbt-labs/dbt-core/blob/main/core/setup.py).

Those requirements in the setup.py file specify all the packages that will be downloaded when you install dbt-core, which means sqlparse will automatically be downloaded with the installation of dbt-core. 

Regardless, our dev_requirements do not impact our adapter's functionality, only development (such as testing).  Here is an example of snowflake's dev requirements: https://github.com/dbt-labs/dbt-snowflake/blob/main/dev-requirements.txt

ArgusLi:
You raise a good point, a lot of the dependencies are taken from dbt-core. We should try and remove as many dependencies from dev_requirements as possible.

Issue #203, opened 2023-08-15
[Bug]: Dots in the schema name do not parse correctly


dremio-brock:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using a dot (.) in a schema name, dbt will parse this and split out the schema to read:
"samples"."dremio"."com"
instead of
"samples.dremio.com"

### Expected Behavior

dbt should parse schema as: "samples.dremio.com"

### Steps To Reproduce

Create a schema.yaml file containing the following

```
sources: 
  - name: samples
    database: Samples
    schema: samples.dremio.com
    tables:
      - name: NYC-weather.csv
      - name: NYC-taxi-trips-iceberg
```

### Environment

```markdown
- OS: Windows 10 / Mac OSX
- dbt-dremio: commit: 388f540fecde5706d11364caf18937553d31ae13
- Dremio Software: Enterprise 24.1
- Dremio Cloud: 8/15/23
```

dremio-brock:
To fix this issue:

Replace [Line 50](https://github.com/dremio/dbt-dremio/blob/388f540fecde5706d11364caf18937553d31ae13/dbt/adapters/dremio/relation.py#L50) with:
```
PATTERN = re.compile(r'''((?:[^."']|"[^"]*"|'[^']*')+)''') 
return ".".join(PATTERN.split(identifier)[1::2])
```

Create your schema using double quotes when specifying a single space/folder and single quotes around the entire schema like the following:

```
sources:
  - name: weather
    database: Samples
    schema: '"samples.dremio.com"'
    tables:
      - name: NYC-weather.csv
  - name: nyc
    database: Samples
    schema: '"samples.dremio.com"."Dremio University"'
    tables:
      - name: sample_data
```

lethalazo:
could you open a PR for the fix you suggested please

PR #205, opened 2023-09-07
Fix schema parsing

ravjotbrar:
### Summary

Prior to this change, users would not be able to specify a schema with a dot in it. For instance, schema.parse would get quoted as "schema"."parse", even though the user intended to quote the entire identifier. Now users can choose to wrap the schema in single quotes and add double quotes around the parts they want. 

### Description

Add regex to match a pattern containing quotes. Solution provided by @dremio-brock. 

### Test Results

All tests pass.

### Changelog

### Related Issue

https://github.com/dremio/dbt-dremio/issues/203


lethalazo (approved):

arseny-dremio (approved):
lgtm

PR #210, opened 2023-11-27
Add version v to macro ref

nigl:
### Summary

Add optional named parameter to macro ref.

### Description

To use model versions with dbt (https://docs.getdbt.com/docs/collaborate/govern/model-versions#how-is-this-different-from-just-creating-a-new-model) it is necessary to have a optional parameter v in the ref macro.

### Test Results

I tested it locally on our dremio installation

### Changelog

-   [ ] Added optional parameter v to the ref macro


ravjotbrar (approved):
LGMT

Issue #176, opened 2023-04-11
[Bug]: Execution time idles for 60 seconds after single-model execution


awaschick:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I execute my (very simple) test model from the dbt command line, execution claims to have taken 62 seconds: 
![image](https://user-images.githubusercontent.com/6818075/231212262-0fd0d363-4f1f-4d23-b931-1d7fd8d62113.png)

However, this does not reflect the actual performance of the queries in Dremio: 
![image](https://user-images.githubusercontent.com/6818075/231212611-9bf656ef-b7a7-44d2-9545-d8b467ae1d07.png)

As you can see, the actual query execution time was less than 4 seconds for everything. Then, dbt apparently cooled its heels for 60 seconds until it hit, presumably, some sort of timeout.  

This is a new behavior from my experience using the (admittedly very old) previous versions of the dbt-dremio adapter.  I understand this new version is making requests in a very different way and I am very happy with all of the new features, but I don't understand why this is happening. 

### Expected Behavior

I expect the model queries to execute and the job to return with completed status immediately afterward. 

### Steps To Reproduce

- Build single dbt model
- Single thread profile or multi-thread profile doesn't matter
- Object storage is not `$scratch`, but local storage
- Run model against Dremio instance

### Environment

```markdown
- OS: Ubuntu in JupyterHub/Kubernetes
- dbt-dremio: 1.4.5
- Dremio Software: 24.0.0 on-premise
- Dremio Cloud: no
```

### Relevant log output

```shell
[15:24:26] jovyan@jupyter-awaschick-40hlkagency-2ecom:~/shared/awaschick/jupyter-automation-qa-delivery-kit/notebooks/dbt/report $ dbt run --vars '{"dremio_database": "_sandbox", "schema_prefix": "hello"}'
15:24:33  Running with dbt=1.4.5
15:24:35  Found 1 model, 0 tests, 0 snapshots, 0 analyses, 333 macros, 0 operations, 0 seed files, 1 source, 0 exposures, 0 metrics
15:24:35  
15:24:38  Concurrency: 2 threads (target='default')
15:24:38  
15:24:38  1 of 1 START sql table model local_nfs_storage_0.dbt-scratch.qa-delivery-kit-dev.hello.hello  [RUN]
[...]

PR #212, opened 2024-01-19
disabled local data fetch for CTAS & limited local data fetch to 100k…

mxmarg:
See description in Dremio-internal JIRA ticket: https://dremio.atlassian.net/browse/DX-86623

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
-# dbt-dremio vNext - release TBD
+# dbt-dremio v1.5.1
 
-## Under the Hood
+## Changes
 
 -   [#199](https://github.com/dremio/dbt-dremio/issues/199) Populate PyPI's `long_description` with contents of `README.md`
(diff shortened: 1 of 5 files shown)


mxmarg:
Hi @awaschick, @Conq1, @ravjotbrar,
I encountered a similar issue to the one discussed here (https://github.com/dremio/dbt-dremio/issues/176), when I tried to run incremental materializations for large tables and dbt was taking hours trying to fetch the records with 100 row GET requests.
I proposed a fix to disable data fetch for CTAS queries and generally limit row fetching to the same value that Dremio's UI does (100k).
Any thoughts or criticisms on this (fairly blunt) approach?

Conq1:
I went the way of controlling it myself:
  {% call statement('create_staging_table', fetch_result=False) -%}
    {{ create_table_as(False, source_staging_relation, sql_select) }}
  {%- endcall %}

But might be a good idea to force it for people less familiar with what's going on.

ravjotbrar:
@mxmarg Looking into this. Agree with changing the row limit size to be 500. It looks like we might be able to make fetch false by default in the code. The only time we want to fetch is when we are getting the columns of a table.

Issue #211, opened 2024-01-19
[Bug]: dbt incremental model takes extremely long to finish when run against large table


mxmarg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Description see Dremio-internal JIRA: https://dremio.atlassian.net/browse/DX-86623

### Environment

```markdown
- OS:
- dbt-dremio:
- Dremio Software:
- Dremio Cloud:
```

mxmarg:
https://github.com/dremio/dbt-dremio/pull/212

mxmarg:
Issue has been fixed by https://github.com/dremio/dbt-dremio/pull/214

PR #214, opened 2024-01-25
Ensure adapter does not fetch actual records from materialization

ravjotbrar:
### Summary

During the second run of an incremental materialization, the adapter would try and fetch records from the temporary table it is creating. This made the adapter hang if the table was large enough. 

### Description
Made the fetch configuration false by default. Also changed the run_query macro to use call_statement, as run_query will always fetch the data that is returned by the sql query. Thanks to @mxmarg for the initial iteration of this fix. 

### Related Issue

https://github.com/dremio/dbt-dremio/issues/211

### DIFF
--- a/dbt/adapters/dremio/api/cursor.py
+++ b/dbt/adapters/dremio/api/cursor.py
@@ -77,7 +77,7 @@ class DremioCursor:
         self._initialize()
         self.closed = True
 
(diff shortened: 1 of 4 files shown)


mxmarg (approved):
I have tested these changes on my machine and they correctly suppress data fetching for incremental models.

arseny-dremio (approved):

PR #216, opened 2024-01-27
Make schema parsing backwards compatible

ravjotbrar:
### Summary

We made changes to the schema parsing that allowed for schemas that contained dots. However, this requires users to specify ALL schemas wrapped in single quotes and then double quotes (`'"my_schema"'`), otherwise it won't add any quotes to schemas. With this change users can continue to add their schema as normal (`'my_schema'`).

### Description

Added a condition to check if double quotes are contained in the schema definition. If so, we use the most recent logic. If not we use the regular logic. 

### Changelog

-   [ ] Added a summary of what this PR accomplishes to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # dbt-dremio vNext - release TBD
 
-## Under the Hood
+## Changes
 
 -   [#199](https://github.com/dremio/dbt-dremio/issues/199) Populate PyPI's `long_description` with contents of `README.md`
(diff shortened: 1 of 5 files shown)


mxmarg (approved):
In my testing, the logic still fails, when I provide a purposefully unconventional object storage path, like this:
`object_storage_path='dbt-test."strange.folder.name"'`

But I think it is reasonable to expect users to avoid using such paths in their dbt models.

Other than that, the code now covers all examples that I have tested.

arseny-dremio (approved):

PR #220, opened 2024-03-13
Optimize REST API workload by using Session and slowing down polling loop

chulucninh09:
### Summary

The cursor implementation was using `while True` loop to poll for result from REST API of Dremio. Without throttling, it will jam the server easily, especially when using a high number of thread count. Since most SQL jobs need a few second to run, looping too fast to get the result is unessesary.

Also, in REST client implementation, using `get`, `post`, `delete` directly from `requests` is inefficient because it will create new HTTP session with the server everytime the polling is called.

### Description

1. Use `request.Session` instead of vanila `requests` to utilize connection pooling, so optimize networking performance
2. Add `time.sleep` in `while True` loop to slow down the polling

### Changelog

- [x] Added a summary of what this PR accomplishes to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Changes
 
--   [195](https://github.com/dremio/dbt-dremio/issues/195) Ensure the adapter does not try and create folders in object storage source
+-   [#195](https://github.com/dremio/dbt-dremio/issues/195) Ensure the adapter does not try and create folders in object storage source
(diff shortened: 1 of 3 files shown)


chulucninh09:
Hi @ravjotbrar, can you comment on this PR and make this progress

ravjotbrar:
Apologies for the late response @chulucninh09. I will try to get to this tomorrow.

chulucninh09:
Hi @ravjotbrar, can I get approval on this or any comment for this PR?

ravjotbrar (approved):
LGTM!

PR #221, opened 2024-03-22
fix for DX-87286, added retries

mxmarg:
see comments on internal Jira DX-87286


ravjotbrar (approved):
All functional tests pass. Looks good, thanks!

Issue #196, opened 2023-07-07
Add ability to create and update Dremio wiki and tag entries


mxmarg:
### Describe the enhancement requested

The ability to set Dremio wiki and tag entries is currently missing.

Wiki and tag entries can be added or updated in the Dremio UI or using the following REST API calls:

- https://docs.dremio.com/software/api/catalog/wiki/
- https://docs.dremio.com/software/api/catalog/tag/

### Justification for this enhancement

Automated documentation is a central part of dbt's value add and dbt even has its own concept of tags, which could be transferred to views created in Dremio

https://www.getdbt.com/product/data-documentation/

Issue #86, opened 2022-11-18
[enhancement] propagate model 's doc & tags to dremio dataset wiki & tags


fabrice-etanchaud:
Hi all ! What about using a model's doc & tags to generate the wiki & tags of the related dremio dataset ?

jlarue26:
Thanks @fabrice-etanchaud ! Not a bad idea!

PR #250, opened 2024-12-03
Implementation of wikis and tags feature

99Lys:
### Summary
Wikis and tags feature implementation.

### Description
One of dbt's core features is source documentation. Right now, this documentation does not propagate into Dremio’s wiki, leading to two separate sources of truth.

The goal is to display dbt documentation - more specifically, wikis and tags - inside of Dremio.

### Test Results
All added tests are running
- Added tests to check wikis and tags creation for table models
- Added tests to check wikis and tags creation, update and deletion for view models

### Changelog

- Possibility to integrate wikis and tags by enabling `relation` option from `persist_docs` configuration
  - New macro `dremio__persist_docs` created
  - Views also perform `persist_docs` macro
  - Integration via REST API

### Related Issue
#86 
#196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
   - `reflections_enabled` adapter option has been renamed to `reflections_metadata_enabled` (requires user privileges to run in dremio)
 -   Removing duplicated macros array_append, array_concat as Dremio already has SQL functions analogues.
 - [#250](https://github.com/dremio/dbt-dremio/pull/250) Possibility to integrate wikis and tags by enabling `relation` option from `persist_docs` configuration
+  - New macro `dremio__persist_docs` created
+  - Views also perform `persist_docs` macro
+  - Integration via REST API
 
 ## Dependency
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Changes
 
 -   Added [DremioRestClient](dbt/adapters/dremio/api/rest/client.py) to isolate all Dremio API calls inside one class
+<<<<<<< HEAD
 - [#256](https://github.com/dremio/dbt-dremio/pull/256) Reflections are now handled through the Rest API
   - Non-admin users are now able to use reflections
   - It is now possible to set a custom name for reflections
@@ -15,6 +16,9 @@
   - Computations default to `SUM, COUNT` if mapped measure is numeric, `COUNT` if not
   - `reflections_enabled` adapter option has been renamed to `reflections_metadata_enabled` (requires user privileges to run in dremio)
 -   Removing duplicated macros array_append, array_concat as Dremio already has SQL functions analogues.
+=======
+
+>>>>>>> 2a84e48 (Wikis and tags feature implementation)
 ## Dependency
 
 -   [#222](https://github.com/dremio/dbt-dremio/issues/222) Upgrade dbt-core to 1.8.8 and dbt-tests-adapter to 1.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## Changes
 
 -   Added [DremioRestClient](dbt/adapters/dremio/api/rest/client.py) to isolate all Dremio API calls inside one class
-<<<<<<< HEAD
 - [#256](https://github.com/dremio/dbt-dremio/pull/256) Reflections are now handled through the Rest API
   - Non-admin users are now able to use reflections
   - It is now possible to set a custom name for reflections
@@ -16,9 +15,8 @@
   - Computations default to `SUM, COUNT` if mapped measure is numeric, `COUNT` if not
   - `reflections_enabled` adapter option has been renamed to `reflections_metadata_enabled` (requires user privileges to run in dremio)
 -   Removing duplicated macros array_append, array_concat as Dremio already has SQL functions analogues.
-=======
+- [#250](https://github.com/dremio/dbt-dremio/pull/250) Possibility to integrate wikis and tags by enabling `relation` option from `persist_docs` configuration
 
->>>>>>> 2a84e48 (Wikis and tags feature implementation)
 ## Dependency
 
 -   [#222](https://github.com/dremio/dbt-dremio/issues/222) Upgrade dbt-core to 1.8.8 and dbt-tests-adapter to 1.8.0
@@ -27,6 +25,7 @@
 
 -   [#223](https://github.com/dremio/dbt-dremio/issues/224) Implement merge strategy for incremental materializations
 -   [#229](https://github.com/dremio/dbt-dremio/issues/229) Add max operator to get_relation_last_modified macro 
+-   [#250](https://github.com/dremio/dbt-dremio/pull/250) Implementation of wikis and tags feature
 
 # dbt-dremio v1.7.0
 
--- a/dbt/adapters/dremio/api/rest/client.py
+++ b/dbt/adapters/dremio/api/rest/client.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 
-
 import requests
 
 from dbt.adapters.dremio.api.authentication import DremioPatAuthentication
--- a/dbt/adapters/dremio/api/rest/client.py
+++ b/dbt/adapters/dremio/api/rest/client.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 
+
 import requests
 
 from dbt.adapters.dremio.api.authentication import DremioPatAuthentication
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -115,4 +118,4 @@
 
 -   Upgrade dbt-core to 1.3.2.
 
--   Upgrade dbt-tests-adapter to 1.3.2.
+-   Upgrade dbt-tests-adapter to 1.3.2.
```
howareyouman, line 121: Could you please keep the empty line in the end of the file? It should be a parameter in your IDE somewhere.

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,79 @@ def create_catalog(self, relation):
...
+            return
+
+        object_id = catalog_info.get("id")
+        stored_wiki = rest_client.retrieve_wiki(object_id)
+        wiki_content = stored_wiki.get("text")
+        wiki_version = stored_wiki.get("version", None)
+
+        if wiki_version == None:
+            logger.debug(f"Creating wiki for {'.'.join(path)}")
+            logger.debug(rest_client.create_wiki(object_id, text))
```
howareyouman, line 262: It's better to separate the call and logging smth. It will be better in visibility and if the result of the function will vary we will be more flexible with changing the logging messages.

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,79 @@ def create_catalog(self, relation):
...
+        elif wiki_content != text:
+            if text == "": # text is empty, delete wiki
+                logger.debug(f"Deleting wiki for {'.'.join(path)}")
+                logger.debug(rest_client.delete_wiki(object_id, wiki_version))
```
howareyouman, line 266: The same  comment about separation of result and logging goes everywhere

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,79 @@ def create_catalog(self, relation):
...
+            else:
+                logger.debug(f"Updating wiki for {'.'.join(path)}")
+                logger.debug(rest_client.update_wiki(object_id, text, wiki_version))
+
+    def docs_integration_with_tags(self, relation, tags: Union[str,list[str]]):
```
howareyouman, line 271: Could you please create objects with the fields, it's not clear from the reviewer and next developers perspective to understand what `Union[str,list[str]]` mean

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,79 @@ def create_catalog(self, relation):
...
+            return
+
+        object_id = catalog_info.get("id")
+        stored_tags = rest_client.retrieve_tags(object_id)
+        tags_list = stored_tags.get("tags")
+        tags_version = stored_tags.get("version", None)
+
+        if tags_version == None:
+            logger.debug(f"Creating tags for {'.'.join(path)}")
+            logger.debug(rest_client.create_tags(object_id, tags))
```
howareyouman, line 298: In general have else if is not the best practice. 
Sometimes it's got to have return after one If and there will be no need to have one else. 

The same recommendation may come to the previous method.

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,79 @@ def create_catalog(self, relation):
             logger.debug(f"Creating folder(s): {database}.{schema}")
             self._create_folders(database, schema, rest_client)
         return
+    
+    # dbt docs integration with Dremio wikis and tags
+    def docs_integration_with_wikis(self, relation, text: str):
```
howareyouman, line 235: All functions should be name as verbs. 
Let's name this function as a  process_wikis()

dbt/adapters/dremio/connections.py
howareyouman, line 271: And let's name this function as process_tags()

dbt/include/dremio/macros/adapters/persist_docs.sql
```diff
@@ -0,0 +1,20 @@
+-- {% macro persist_docs(relation, model, for_relation=true, for_columns=true) -%}
```
howareyouman, line 1: Do we need all these rows commented? Could these lines be deleted from this file?

tests/functional/adapter/dremio_specific/test_persist_docs.py
```diff
@@ -0,0 +1,352 @@
...
+
+    def test_table_wikis_and_tags(self, project, client):
+        run_dbt(["run", "--select", "table_model"])
+        object_id = self._get_relation_id(project, client, "table_model")
+        wiki = client.retrieve_wiki(object_id)
+        tags = client.retrieve_tags(object_id)
+        self._assert_table_wikis_and_tags(wiki, tags)
+
+    def test_view_wikis_and_tags(self, project, client, models):
+        # Create + Get
```
howareyouman, line 286: Is it possible to have 1 verification in 1 test function? 
It's not great to run multiple verifications in 1 function, because usually 1 test function means 1 functionality test.

We have no limitations in creating as many functions in the class as needed

howareyouman (commented):
It's great that you Liliana did it yourself!
All the comments I have are more about style and naming.
I don't see any huge blockers to get it merged soon!

dbt/adapters/dremio/connections.py
99Lys: In the beginning I put this because `tags` can be configured in the YAML as a `string` (one tag) or a `list of strings` (many tags), and that's why I used `Union` here - to express tags can be `str` or `list[str]`. But then I realized that actually even if the configured tag is just a string (in the .yml), internally dbt handles it as a list of strings with one element. 

I forgot to remove this type, will fix it to be `list[str]` only.

tests/functional/adapter/dremio_specific/test_persist_docs.py
99Lys: Actually the way I test the successful creation of wikis and tags are by getting the values (so I use `retrieve_wikis` function) and see if they are the supposed ones. So the getting wikis/tags methods are not actually being tested, they seem to be dependent of creation, vice-versa.

Could you please suggest how could I test them separately?

(I already created separated functions for `update` and `delete`)

bcmeireles (commented):
Great PR!
@howareyouman already pointed pretty much everything, I just have a few questions where I might be lacking context to understand

tests/functional/adapter/dremio_specific/test_persist_docs.py
```diff
@@ -0,0 +1,361 @@
...
+        profile["test"]["outputs"]["default"] = target
+
+        if profiles_config_update:
+            profile.update(profiles_config_update)
+        return profile
+    
+    @pytest.fixture(scope="class", autouse=False)
+    def setUp(self, project):
+        run_dbt(["seed"])
+        run_dbt()
```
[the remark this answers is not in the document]
bcmeireles: Asking as from what I know `dbt seed` is used to load csv files from `/seeds` but I don't see any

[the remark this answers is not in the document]
99Lys: This `setUp` function originally exists in `BasePersistDocsBase` class (from where `BasePersistDocs` is inherited). This is a fixture set to be running automatically `@pytest.fixture(scope="class", autouse=True)` in `autouse=True`. Since I wanted to have control on when to run dbt, I just copy pasted it and changed the annotation to `autouse=False`. 
I did something similar for function `test_has_comments_pglike` by replacing its content so this test is not being run.
I don't know what is the best practice here, please let me know if there is a best way to do this!  Thanks

dbt/adapters/dremio/api/rest/client.py
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 
+
```
simonpannek, line 18: [nit]
undo

dbt/adapters/dremio/connections.py
```diff
@@ -231,6 +230,92 @@ def create_catalog(self, relation):
...
+        identifier = relation.identifier
+        path.append(identifier)
+        try:
+            catalog_info = rest_client.get_catalog_item(
+                catalog_id=None,
+                catalog_path=path,
+            )
+        except DremioNotFoundException:
+            logger.debug("Catalog not found. Returning")
+            return
```
simonpannek, line 295: Is it fine for us just to return here? Should dbt error here?

tests/functional/adapter/dremio_specific/test_persist_docs.py
```diff
@@ -0,0 +1,361 @@
...
+
+        catalog_info = client.get_catalog_item(
+            catalog_id=None,
+            catalog_path=path,
+        )
+        return catalog_info.get("id")
+    
+    # Overriding the original test, to be ignored
+    def test_has_comments_pglike(self, project):
+        pass
```
simonpannek, line 276: Could you add a comment on why we are overriding this test?

simonpannek (dismissed):
Good job! No blockers on my side, just a few small comments!

tests/functional/adapter/dremio_specific/test_persist_docs.py
```diff
@@ -0,0 +1,361 @@
...
+
+        return client
+    
+    # Removing unnecessary models and adding schema
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "table_model.sql": _MODELS__TABLE,
+            "view_model.sql": _MODELS__VIEW,
+            "schema.yml": _PROPERTIES__SCHEMA_YML,
```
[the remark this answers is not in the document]
99Lys: From what I understood, `unique_schema` is only "building" the correct path where the models are being created and that is later used in `dbt_profile_data` to update the project configuration. 
Initially, I was setting `schema.yml` in properties. But since I needed to update it later, I moved to models because `properties/` doesn't seem to exist so I could only perform `write_file(<updated_schema.yml>, project.project_root, "models", "schema.yml")`.
Also I don't think it's a problem to configure `schema.yml` in models because in an actual project by default there is where it is stored.

tests/functional/adapter/dremio_specific/test_persist_docs.py
howareyouman: I see that you already have it, so I'm closing this conversation!

tests/functional/adapter/dremio_specific/test_persist_docs.py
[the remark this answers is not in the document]
howareyouman: Could we chain dbt commands here? seed and run in 1 command?

howareyouman (dismissed):
Could you please also answer and finish all the discussions we have here? 
Great PR!

dbt/adapters/dremio/connections.py
99Lys: In this case I followed the way of implementation of `drop_catalog` method, and I think no dbt error should be raised here because technically this is not related with dbt itself and doesn't affect the normal execution of dbt

tests/functional/adapter/dremio_specific/test_persist_docs.py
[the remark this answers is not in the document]
99Lys: I don't think so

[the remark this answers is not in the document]
howareyouman: @99Lys If you will take a look on the code of run_dbt() function you will take a look that it's uses command run. So to chain operations in this function you can easily run run_dbt(["seed", "run"]) and that will be 1 dbt command run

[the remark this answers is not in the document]
99Lys: Decided as a team - since this function is not used in this feature's tests at all, besides updating the fixture annotation to `autouse=False`, I will also remove the method body to clearly transmit this idea.
Improving tests in general will be done in other issue that will be created.

howareyouman (approved):
Awesome!

Issue #184, opened 2023-06-02
Add depend strategy for reflections


darian-heede:
### Describe the enhancement requested

Currently, when creating a reflection, dbt-dremio triggers it on dremio and instantly finishes the job if the trigger was successful. We would like to be able to configure a strategy that makes dbt wait for the completion of the reflection job on dremio, before marking it as successful and moving on to the next item in the dbt DAG. This would also allow us to get feedback on a reflection failure.

The configuration keyword could be `strategy` with the possible values `depend` or `trigger`, the later being the current and default value.

### Justification for this enhancement
This feature would allow for better control and debugging of reflection jobs.

darian-heede:
Hi there, any feedback for this issue?

PR #256, opened 2024-12-19
Move reflections to rest API calls

bcmeireles:
### Summary

Reflections were being created through the SQL Runner and that was the cause for some issues we were facing.

### Description

The current reflections implementation requires sys.reflections access, which is only available to superusers (admins). When creating the reflection with a regular user, it will error. To fix this issue, the creation of the reflections has been moved from queries being ran inside the SQL Runner to Dremio's Rest API calls.
The current logic of writing reflections through a custom materialization remains to make sure reflections created under the previous approach still work.

On top of this, the behaviour where all reflections would be dropped and then recreated when running a dbt project has been changed to instead first check if a reflection with the same name is already present in the same dataset and, if that's the case, it will update the already existing reflection instead. 

### Test Results

wip

### Changelog

- Reflections are now handled through the Rest API, allowing non-admin users to also create them
- It is now possible to set a custom name for reflections
- If a reflection already exists in the dataset with the same name defined in the model, it will be updated instead of creating a new one
- New `date_dimensions` parameter was added to the reflection materialization, to set fields that have a `DATE` granularity

### DIFF
--- a/dbt/adapters/dremio/connections.py
+++ b/dbt/adapters/dremio/connections.py
@@ -30,6 +30,9 @@ from dbt.adapters.sql import SQLConnectionManager
 from dbt.adapters.contracts.connection import AdapterResponse
 
 from dbt.adapters.dremio.api.rest.endpoints import (
+    create_reflection,
+    update_reflection,
+    get_reflection,
     delete_catalog,
     create_catalog_api,
     get_catalog_item,
@@ -133,7 +136,7 @@ class DremioConnectionManager(SQLConnectionManager):
 
     # Auto_begin may not be relevant with the rest_api
     def add_query(
-        self, sql, auto_begin=True, bindings=None, abridge_sql_log=False, fetch=False
+            self, sql, auto_begin=True, bindings=None, abridge_sql_log=False, fetch=False
     ):
         connection = self.get_thread_connection()
         if auto_begin and connection.transaction_open is False:
@@ -174,11 +177,11 @@ class DremioConnectionManager(SQLConnectionManager):
         return AdapterResponse(_message=message, rows_affected=rows)
 
     def execute(
-        self,
-        sql: str,
-        auto_begin: bool = False,
-        fetch: bool = False,
-        limit: Optional[int] = None,
+            self,
+            sql: str,
+            auto_begin: bool = False,
+            fetch: bool = False,
+            limit: Optional[int] = None,
     ) -> Tuple[AdapterResponse, agate.Table]:
         sql = self._add_query_comment(sql)
         _, cursor = self.add_query(sql, auto_begin, fetch=fetch)
@@ -231,6 +234,76 @@ class DremioConnectionManager(SQLConnectionManager):
             self._create_folders(database, schema, api_parameters)
         return
 
+    def dbt_reflection_integration(self, name: str, type: str, anchor, display, dimensions, date_dimensions, measures,
+                                   computations, partition_by, partition_method, localsort_by):
+        thread_connection = self.get_thread_connection()
+        connection = self.open(thread_connection)
+        api_parameters = connection.handle.get_parameters()
+
+        database = anchor.database
+        schema = anchor.schema
+        path = self._create_path_list(database, schema)
+        identifier = anchor.identifier
+
+        path.append(identifier)
+
+        catalog_info = get_catalog_item(
+            api_parameters,
+            catalog_id=None,
+            catalog_path=path,
+        )
+
+        dataset_id = catalog_info.get("id")
+
+        payload = {
+            "type": type,
+            "name": name,
+            "datasetId": dataset_id,
+            "enabled": True,
+            "arrowCachingEnabled": False,
+            "partitionDistributionStrategy": partition_method.upper(),
+            "entityType": "reflection"
+        }
+
+        if display:
+            payload["displayFields"] = [{"name": field} for field in display]
+
+        if dimensions:
+            if not date_dimensions:
+                date_dimensions = []
+
+            payload["dimensionFields"] = [
+                {"name": dimension} if dimension not in date_dimensions else {"name": dimension, "granularity": "DATE"}
+                for dimension in dimensions]
+
+        if measures and computations:
+            payload["measureFields"] = [{"name": measure, "measureTypeList": computation.split(',')} for
+                                        measure, computation in zip(measures, computations)]
+
+        if partition_by:
+            payload["partitionFields"] = [{"name": partition} for partition in partition_by]
+
+        if localsort_by:
+            payload["sortFields"] = [{"name": sort} for sort in localsort_by]
+
+        dataset_info = get_reflection(api_parameters, dataset_id)
+        reflections_info = dataset_info.get("data")
+
+        updated = False
+        for reflection in reflections_info:
+            if reflection.get("name") == name:
+                logger.debug(f"Reflection {name} already exists. Updating it")
+                payload["tag"] = reflection.get("tag")
+                logger.info(
+                    update_reflection(api_parameters, reflection.get("id"), payload))
+                updated = True
+                break
+
+        if not updated:
+            logger.debug(f"Reflection {name} does not exist. Creating it")
+            logger.info(
+                create_reflection(api_parameters, name, type, payload))
+
     def _make_new_space_json(self, name) -> json:
         python_dict = {"entityType": "space", "name": name}
         return json.dumps(python_dict)
@@ -263,6 +336,7 @@ class DremioConnectionManager(SQLConnectionManager):
 
     def _create_path_list(self, database, schema):
         path = [database]
-        folders = schema.split(".")
-        path.extend(folders)
+        if schema != 'no_schema':
+            folders = schema.split(".")
+            path.extend(folders)
         return path
--- a/dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
+++ b/dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
@@ -37,85 +37,18 @@ ADD EXTERNAL REFLECTION name
 USING target
 #}
 
-{%- macro create_reflection(reflection_type, anchor, reflection, external_target=none,
-  display=none, dimensions=none, by_day_dimensions=none, measures=none) %}
-  alter dataset {{ anchor }}
-    create {{ reflection_type }} reflection {{ reflection.include(database=False, schema=False) }}
-    using
-    {%- if reflection_type == 'raw' %}
-      {{ display_clause(display) }}
-    {%- elif reflection_type == 'aggregate' %}
-      {{ dimensions_clause(dimensions=dimensions, by_day_dimensions=by_day_dimensions) }}
-      {{ measures_clause(measures) }}
-    {%- else -%}
-      {{ external_target }}
-    {% endif -%}
-    {%- if reflection_type in ['raw', 'aggregate'] %}
-      {{ partition_method() }} {{ config_cols("partition by") }}
-      {{ config_cols("localsort by") }}
-      {{ config_cols("distribute by") }}
-      {{ arrow_cache_clause() }}
-    {%- endif -%}
-{% endmacro -%}
+{%- macro create_reflection(reflection_name, reflection_type, anchor,
+  display=none, dimensions=none, date_dimensions=none, measures=none, computations=none, partition_by=none, partition_method=none, localsort_by=none) %}
 
-{%- macro display_clause(display=none) %}
-  {%- set cols = config.get('display', validator=validation.any[list, basestring]) or display -%}
-  {%- if cols is not none %}
-    {%- if cols is string -%}
-      {%- set cols = [cols] -%}
-    {%- endif -%}
-    display (
-    {%- for item in cols -%}
-      {{ adapter.quote(item) }}
-      {%- if not loop.last -%},{%- endif -%}
-    {%- endfor -%}
-    )
-  {%- endif %}
-{% endmacro -%}
-
-{%- macro dimensions_clause(dimensions=none, by_day_dimensions=none) %}
-  {%- set cols = config.get('dimensions', validator=validation.any[list, basestring]) or dimensions -%}
-  {%- set by_day_cols = config.get('by_day_dimensions', validator=validation.any[list, basestring]) or by_day_dimensions -%}
-  {%- if cols is not none %}
-    {%- if cols is string -%}
-      {%- set cols = [cols] -%}
-    {%- endif -%}
-    {%- if by_day_cols is string -%}
-      {%- set by_day_cols = [by_day_cols] -%}
-    {%- endif -%}
-    dimensions (
-    {%- for item in cols -%}
-      {{ adapter.quote(item) ~ (' by day' if item in by_day_cols else "") }}
-      {%- if not loop.last -%},{%- endif -%}
-    {%- endfor -%}
-    )
-  {%- endif %}
-{% endmacro -%}
-
-{%- macro measures_clause(measures=none) %}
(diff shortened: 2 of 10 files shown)


.github/scripts/create_and_format_sources.sh
```diff
@@ -50,4 +50,28 @@ curl -s -X PUT "http://localhost:9047/apiv2/source/dbt_test_source" \
...
+  -H "Authorization: _dremio$AUTH_TOKEN" \
+  --data-raw "{\"type\":\"JSON\"}"
+
+echo "SF_incidents2016 formatted in Dremio."
+
+echo "Formatting NYC-taxi-trips..."
+curl -s -X PUT "http://localhost:9047/apiv2/source/Samples/folder_format/samples.dremio.com/NYC-taxi-trips" \
+  -H "Content-Type: application/json" \
+  -H "Authorization: _dremio$AUTH_TOKEN" \
+  --data-raw "{\"ignoreOtherFileFormats\":false,\"type\":\"Parquet\"}"
```
howareyouman, line 75: Should it be Iceberg format? Or this only a parquet?

dbt/adapters/dremio/connections.py
```diff
@@ -230,6 +231,102 @@ def create_catalog(self, relation):
...
+        path.append(identifier)
+
+        catalog_info = rest_client.get_catalog_item(
+            catalog_id=None,
+            catalog_path=path,
+        )
+
+        dataset_id = catalog_info.get("id")
+
+        payload = {
```
howareyouman, line 257: Let's start creating entities for requests with fields inside it.
And make them a part of the rest client.

dbt/adapters/dremio/connections.py
```diff
@@ -230,6 +231,102 @@ def create_catalog(self, relation):
...
+        if partition_by:
+            if not partition_transform:
+                partition_transform = ["IDENTITY"] * len(partition_by)
+
+            partition_fields = []
+            for partition, transform in zip(partition_by, partition_transform):
+                transform = transform.upper()
+                partition_field = {"name": partition, "transform": None}
+
+                if transform in ["YEAR", "MONTH", "DAY", "HOUR", "IDENTITY"]:
```
howareyouman, line 290: These strings should be an enum

dbt/adapters/dremio/connections.py
```diff
@@ -230,6 +231,102 @@ def create_catalog(self, relation):
...
+                    partition_field["transform"] = {"type": transform}
+                elif transform.startswith("BUCKET"):
+                    bucket_count = int(transform.split("(")[1].split(")")[0])
```
howareyouman, line 293: Could you please describe this format somewhere in the docs + create a separate function where you will be parsing it

dbt/adapters/dremio/connections.py
```diff
@@ -263,6 +360,7 @@ def _create_folders(self, database, schema, rest_client: DremioRestClient):
 
     def _create_path_list(self, database, schema):
         path = [database]
-        folders = schema.split(".")
-        path.extend(folders)
+        if schema != 'no_schema':
```
howareyouman, line 363: Why do we need this?

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
```diff
@@ -37,85 +37,18 @@ ADD EXTERNAL REFLECTION name
 USING target
 #}
 
-{%- macro create_reflection(reflection_type, anchor, reflection, external_target=none,
```
howareyouman, line 40: Let's start creating proper documentation about it. 
This is a blocking change for customers, it will be 100% a minor release of the dbt connector.
And the users of it should know what happened and why.

dbt/include/dremio/macros/materializations/reflection/reflection.sql
```diff
@@ -60,28 +74,48 @@ limitations under the License.*/
...
+    {% elif reflection_type in ['aggregate', 'aggregation'] %}
       {% if dimensions is none %}
         {% set dimensions = columns | rejectattr('dtype', 'in', ['decimal', 'float', 'double']) | map(attribute='name') | list %}
-        {% set by_day_dimensions = columns | selectattr('dtype', 'in', ['timestamp']) | map(attribute='name') | list %}
+        {% set date_dimensions = columns | selectattr('dtype', 'in', ['timestamp']) | map(attribute='name') | list %}
       {% endif %}
       {% if measures is none %}
         {% set measures = columns | selectattr('dtype', 'in', ['decimal', 'float', 'double']) | map(attribute='name') | list %}
       {% endif %}
+      {% if computations is none %}
```
howareyouman, line 89: Is this a support of more complex fields in Dremio rest API of reflections?

tests/functional/adapter/dremio_specific/test_reflections.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def _create_path_list(self, database, schema):
+        path = [database]
+        if schema != 'no_schema':
+            folders = schema.split(".")
+            path.extend(folders)
+        return path
+
+    def _get_reflection(self, project, client, identifier, expected_name):
+        client.start()
```
howareyouman, line 165: Should we start the client in `client` method? and rename it to `get_client`?

howareyouman (commented):
Great PR! 
I have some not huge blocking changes requests.

The main part is documentation. We should update changelog, we should create a ticket for docs team for sure about this feature in our dbt connector.

I remember that we were discussing the waiting part of reflection to be created. Will we provide it in the next PR? Should we create a ticket for it?

howareyouman:
@bcmeireles we should transfer the PR description somewhere in the docs and changelog as well!

dbt/adapters/dremio/connections.py
bcmeireles: Before a no_schema's path would be evaluate to `[[@normal, no_schema]]` which is incorrect as by having no schema means that in reality we want `['@normal']`

.github/scripts/create_and_format_sources.sh
bcmeireles: Only a parquet, we are using `NYC-taxi-trips` not `NYC-taxi-trips-iceberg`

dbt/include/dremio/macros/materializations/reflection/reflection.sql
bcmeireles: Yes

bcmeireles: More specifically, `measureTypeList` parameter of [measureFields](https://docs.dremio.com/24.3.x/reference/api/reflections/#parameters-of-the-measurefields-object)

dbt/adapters/dremio/connections.py
bcmeireles: Will be passing it to a different function, I plan to update [the Reflections section of the wiki](https://github.com/dremio/dbt-dremio/wiki/Using-Materializations-with-Dremio#reflections) once merged

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
bcmeireles: The changes will be documented in the wiki; besides `date_dimensions` no functionality was changed, only added

bcmeireles:
> Great PR! I have some not huge blocking changes requests.
> 
> The main part is documentation. We should update changelog, we should create a ticket for docs team for sure about this feature in our dbt connector.
> 
> I remember that we were discussing the waiting part of reflection to be created. Will we provide it in the next PR? Should we create a ticket for it?

> @bcmeireles we should transfer the PR description somewhere in the docs and changelog as well!

I'll be copying the PR description into the Changelog. Once merged I will also be updating this repo's wiki with the new functionalities and changes.
Regarding the waiting for a reflection, that has been requested in #184 and is not addressed in this PR.

tests/functional/adapter/dremio_specific/test_reflections.py
bcmeireles: Would we just login as soon as the client is initialized then? Because currently the only thing the `start()` method in the client really does [is call `__login()`](https://github.com/dremio/dbt-dremio/blob/133108cb520e75db0f6dce138ea872f3e2ec05a4/dbt/adapters/dremio/api/rest/client.py#L36-L41)

howareyouman: > Would we just login as soon as the client is initialized then?
That's a good question to discuss. We should think about using the same client in multiple places and ideally this start method should be a part of constructor, but now I decided to leave it only started in handle method.

We can discuss it with @simonpannek as well. It would be great to gather our opinions and make a decision about it.

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
```diff
@@ -37,85 +37,18 @@ ADD EXTERNAL REFLECTION name
...
-    {%- endfor -%}
-    )
-  {%- endif %}
-{% endmacro -%}
+{%- macro create_reflection(reflection_name, reflection_type, anchor,
+  display=none, dimensions=none, date_dimensions=none, measures=none, computations=none, partition_by=none, partition_transform=none, partition_method=none, distribute_by=none, localsort_by=none, arrow_cache=false) %}
+
+  {%- if reflection_type == 'raw' %}
+    {% set reflection_type = 'RAW' %}
+  {%- elif reflection_type in ['aggregate', 'aggregation'] %}
```
simonpannek, line 45: I understand that you're including the duplicate check for both `aggregate` and `aggregation` for backwards compatibility, but since this needs to be done in a lot of different places, this seems very error prone.
We should either abstract this or just stick with `aggregate` only.

dbt/include/dremio/macros/materializations/reflection/reflection.sql
```diff
@@ -60,28 +74,48 @@ limitations under the License.*/
...
+        {% if measures is not none %}
+          {% set computations = [] %}
+          {% for measure in measures %}
+            {% if measure in columns | selectattr('dtype', 'in', ['decimal', 'float', 'double']) | map(attribute='name') | list %}
```
simonpannek, line 93: Can we link to Dremio's documentation on supported computations/type?

simonpannek (changes requested):

dbt/include/dremio/macros/materializations/reflection/reflection.sql
bcmeireles: It's the `measureTypeList` parameter of the [`measureFields` Object](https://docs.dremio.com/24.3.x/reference/api/reflections/#parameters-of-the-measurefields-object)
You mean linking it inside the macro itself?

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
bcmeireles: I'm not necessarily against keeping `aggregate` only as that's how they currently are defined.

But, I believe that `aggregation` is a more correct way to name it (and would keep the possibility of `aggregate` for backwards compatibility. In here I'd argue that maintaining the few (4) places where this possibility was added is more beneficial than completely renaming it and consequently breaking current models. This wouldn't even need to be something we'd be touching (unless we need another rename later?)

What would an abstraction for this be? Creating a macro like
```
{% macro verify_reflection_type_is_aggregation(reflection_type) %}
  {% if reflection_type in ['aggregate', 'aggregation'] %}
    {% do return(true) %}
  {% endif %}
  {% do return(false) %}
{% endmacro %}
```
And calling it
```
{% if {{ verify_aggregation_type(reflection_type) }} %}
```
Instead of verifying through that list?

dbt/include/dremio/macros/materializations/reflection/reflection.sql
simonpannek: Exactly. So you can understand where this is coming from

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
simonpannek: That would be one way of doing it. Alternatively, you could also create a normalize reflection type function that matches strings to enums or create a macro that returns lists of reflection_types.
I'd just try to avoid having this duplication in a lot of different places, independent from each other.

tests/functional/adapter/dremio_specific/test_reflections.py
bcmeireles: Keeping it as is as having the client initialized but not logged in could come in handy in the future

dbt/include/dremio/macros/materializations/reflection/create_reflection.sql
bcmeireles: Solved by checking for both options only once, and then normalizing it there

dbt/adapters/dremio/connections.py
```diff
@@ -230,6 +232,47 @@ def create_catalog(self, relation):
...
+        path.append(identifier)
+
+        catalog_info = rest_client.get_catalog_item(
+            catalog_id=None,
+            catalog_path=path,
+        )
+
+        dataset_id = catalog_info.get("id")
+
+        payload = ReflectionEntity(name, reflection_type, dataset_id, display, dimensions, date_dimensions, measures, computations, partition_by, partition_transform, partition_method, distribute_by, localsort_by, arrow_cache).build_payload()
```
howareyouman, line 258: Should we use multiple lines to have only 80 symbols in one line?

howareyouman (approved):
Could you please add somewhere links about REST API of reflections in Dremio?
The rest is great, thank you!

dbt/adapters/dremio/connections.py
bcmeireles: Formatted

bcmeireles:
A link to the Reflections API docs has also been added in the entity

simonpannek (dismissed):

dbt/adapters/dremio/connections.py
howareyouman: I don't see any change yet. Did you push the commit?

dbt/adapters/dremio/api/rest/entities/reflection.py
```diff
@@ -0,0 +1,141 @@
...
+            truncate_length = int(raw_str.split("(")[1].split(")")[0])
+            return {
+                "type": "TRUNCATE",
+                "truncateTransform": {"truncateLength": truncate_length},
+            }
+
+        return {"type": TransformType.IDENTITY.value}
+
+
+# https://docs.dremio.com/24.3.x/reference/api/reflections/
```
howareyouman, line 54: Should we have latest version of the docs link? Without 24.3.x version?

dbt/adapters/dremio/connections.py
bcmeireles: fixed

dbt/adapters/dremio/api/rest/entities/reflection.py
howareyouman: Let's use standard python documentation way next time. https://realpython.com/documenting-python-code/
It's good to start describing what do we have in the implementation and why do we have it.

howareyouman (approved):

simonpannek (approved):

mxmarg:
Hi @bcmeireles, I have a question:
I tried out the new reflection functionalities and noticed that it did not seem to make any difference whether I set 
```
vars:
  dremio:reflections_enabled: true
```
or
```
vars:
  dremio:reflections_metadata_enabled: true
```

or nothing at all in my `dbt_project.yaml`. Is this flag even needed for the new behavior to work or is it mainly there for backwards compatibility?

```
09:11:08  Running with dbt=1.9.1
09:11:08  Registered adapter: dremio=1.8.1
```

Issue #177, opened 2023-04-13
Allow grants to work with roles


maxfirman:
### Describe the enhancement requested

Currently the grants only work for users not roles. This is hard coded [here](https://github.com/dremio/dbt-dremio/blob/main/dbt/include/dremio/macros/adapters/apply_grants.sql#L39).

It would be convenient to be able to define grants for both users and roles. 

The 'BigQuery' dbt adapter accomplishes this by allow the grantee to be specified with a role/user prefix separated by a colon. For example `user:user_xyz` or `role:role_abc`. See the [dbt docs](https://docs.getdbt.com/reference/resource-configs/grants#bigquery-examples) for an example.

It would be great to add the same functionality to the Dremio dbt adapter. Grantees specified without a role/user prefix could default to 'user' for backwards compatibility. 

### Justification for this enhancement

Ability to specify Dremio 'role' permissions from DBT.

ravjotbrar:
Thanks for raising this enhancement request @maxfirman!

PR #259, opened 2025-01-03
Support for roles in grants

bcmeireles:
### Summary

Added support for roles in grants

### Description

Grants can now be set for both users and for roles.
A prefix was added to handle this, with `user:` and `role:` being the valid prefixes. For example, `user:dbt_test_user_1` and `role:dbt_test_role_1`.
If no prefix is provided, defaults to user for backwards compatibility.

### Test Results

Tested manually, currently working on tests that will test both user and role grants. Extending the pipeline will be needed again as we will also need to create the roles in the dremio instance used for testing

### Changelog

-  Grants now support roles

### Related Issue

#177

### DIFF
--- a/dbt/include/dremio/macros/adapters/apply_grants.sql
+++ b/dbt/include/dremio/macros/adapters/apply_grants.sql
@@ -16,6 +16,23 @@ limitations under the License.*/
     {{ return(False) }}
 {%- endmacro -%}
 
+{%- macro dremio__split_grantee(grantee) -%}
+    {%- set splitted = grantee.split(':') %}
+    {%- if splitted | length == 2 %}
+        {%- set type = splitted[0] %}
+        {%- if type in ['user', 'role'] %}
+            {%- set name = splitted[1] %}
+        {%- else %}
+            {% do exceptions.CompilationError("Invalid prefix. Use either user or role") %}
+        {%- endif %}
+    {%- else %}
+        {%- set type = 'user' %}
+        {%- set name = grantee %}
+    {%- endif %}
+
+    {{ return((type, name)) }}
+{%- endmacro -%}
+
 {% macro dremio__get_show_grant_sql(relation) %}
     {%- if relation.type == 'table' -%}
         {%- set relation_without_double_quotes = target.datalake ~ '.' ~ target.root_path ~ '.' ~ relation.identifier-%}
@@ -36,7 +53,10 @@ limitations under the License.*/
 {% endmacro %}
 
 {%- macro dremio__get_grant_sql(relation, privilege, grantees) -%}
-    grant {{ privilege }} on {{relation.type}} {{ relation }} to user {{adapter.quote(grantees[0])}}
+    {%- set type, name = dremio__split_grantee(grantees[0]) %}
+
+    grant {{ privilege }} on {{ relation.type }} {{ relation }}
+    to {{ type }} {{ adapter.quote(name) }}
 {%- endmacro -%}
 
 {%- macro default__get_revoke_sql(relation, privilege, grantees) -%}


tests/functional/adapter/grants/test_snapshot_grants.py
```diff
@@ -37,6 +37,7 @@
 """.strip()
 
 
+@pytest.mark.skip(reason="Dremio only supports grants in EE/DC editions.")
```
simonpannek, line 40: Since we're probably going to run test pipelines against EE in the future as well, could we add a [conditional](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-skipif-ref) to these tests?

dbt/include/dremio/macros/adapters/apply_grants.sql
```diff
@@ -16,6 +16,23 @@ limitations under the License.*/
     {{ return(False) }}
 {%- endmacro -%}
 
+{%- macro dremio__split_grantee(grantee) -%}
+    {%- set splitted = grantee.split(':') %}
```
simonpannek, line 20: Double checking, is `:` explicitly not allowed as a character in both user and role names?

simonpannek: I checked myself - for usernames that's indeed the case, but roles can contain the `:` character which would break this solution.

tests/functional/adapter/grants/test_snapshot_grants.py
bcmeireles: These will be ran against EE and DC. We should come back to the skipped tests and add the conditional once we also have the pipelines and know what we should check for

dbt/include/dremio/macros/adapters/apply_grants.sql
bcmeireles: Fixed the bug, also added a test about it

tests/functional/adapter/grants/test_model_grants.py
```diff
@@ -98,7 +226,7 @@ def test_view_table_grants(self, project, get_test_users):
         manifest = get_manifest(project.project_root)
         model = manifest.nodes[model_id]
         assert model.config.materialized == "table"
-        expected = {select_privilege_name: [test_users[1]]}
+        expected = {select_privilege_name: ["user:" + test_users[1]]}
         self.assert_expected_grants_match_actual(project, "my_model", expected)
```
99Lys, line 248: Could we have different tests for each of these assertions (and below)?

99Lys (changes requested):

tests/functional/adapter/grants/test_model_grants.py
bcmeireles: I'm keeping the same structure as we had before and that dbt also uses on their own tests that were the base for it, and this way revokes are also being tested

dbt/include/dremio/macros/adapters/apply_grants.sql
simonpannek: This doesn't fix all edge-cases though. If someone has a legacy system setup with roles that contain colons, execution will fail now since it's trying to match the prefix against `user` or `role`.

simonpannek: Okay, reading through this again, I guess that's not possible since before this PR, grants were only able to target users?

simonpannek: Resolving this since I think this does solve the issue. Feel free to open it again if you disagree.

tests/functional/adapter/grants/test_snapshot_grants.py
simonpannek: Adding the conditional now already would save us some searching through all tests later on, but I won't be blocking on this.

dbt/include/dremio/macros/adapters/apply_grants.sql
```diff
@@ -17,20 +17,20 @@ limitations under the License.*/
...
-    {%- set splitted = grantee.split(':') %}
-    {%- if splitted | length == 2 %}
-        {%- set type = splitted[0] %}
-        {%- if type in ['user', 'role'] %}
-            {%- set name = splitted[1] %}
-        {%- else %}
+    {%- set splitted = grantee.split(':') -%}
+
+    {%- if splitted | length < 2 -%}
+        {{ return(("user", grantee)) }}
```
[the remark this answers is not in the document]
bcmeireles: If others agree it'd add some extra logic but in the end it could be done. It'd involve getting all grants for the specific object (which we also do to determine if we should revoke anything), and then comparing of any role matches the name given by the user.
But at that point, instead of erroring like that we could also just apply it to the role as default behavior in those cases no?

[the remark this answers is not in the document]
simonpannek: I'd avoid adding any "magic" here since this could give us issue in case someone has a username matching a role name.
An easier solution might be to add a deprecation warning in case someone tries to target a user without a prefix?

tests/functional/adapter/grants/test_model_grants.py
bcmeireles: Also since these should override the default dbt tests, we should keep the same structure

dbt/include/dremio/macros/adapters/apply_grants.sql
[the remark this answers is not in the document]
bcmeireles: I'll add such warning 👍

tests/functional/adapter/grants/test_snapshot_grants.py
howareyouman: We could use these custom flags. We should create our own flag for Dremio version and specify it only when it's required, like in this case.

dbt/include/dremio/macros/adapters/apply_grants.sql
```diff
@@ -16,6 +16,23 @@ limitations under the License.*/
     {{ return(False) }}
 {%- endmacro -%}
 
+{%- macro dremio__split_grantee(grantee) -%}
```
howareyouman, line 19: After reading @simonpannek comment, I started thinking about adding validations to all the macros parameters we have. 
Of course, it's not Java and nullability here is not the case, but when we should parse original string, we should have some valid format of it and when the format is incorrect, we should throw an exception.

howareyouman (commented):
It would be great to start adding some validations for macros parameters and add pytest feature flags for test execution.

tests/functional/adapter/grants/test_snapshot_grants.py
bcmeireles: An env variable specifying the Dremio version that is running, and set during the pipeline execution will be added as discussed

dbt/include/dremio/macros/adapters/apply_grants.sql
bcmeireles: https://github.com/dremio/dbt-dremio/blob/53b4dd370f3cd4cd32bbae82d3b1ade38163f47a/dbt/include/dremio/macros/adapters/apply_grants.sql#L28-L29

We have a verification here already in case the user has a prefix that differs from `user` or `role`. If no prefix is provided, we are defaulting back to `user` for backwards compatibility. Following Simon's suggestion, I plan to add a deprecation warning and, in a future version, expand the verification to be more strict for the prefixes. Do you agree?

howareyouman (dismissed):

99Lys (approved):

tests/functional/adapter/grants/test_invalid_grants.py
```diff
@@ -12,13 +12,15 @@
...
 
+import pytest
 
 from dbt.tests.adapter.grants.test_invalid_grants import BaseInvalidGrants
 from tests.functional.adapter.grants.base_grants import BaseGrantsDremio
 from tests.utils.util import relation_from_name
 from dbt.tests.util import get_connection
 
 
+@pytest.mark.skip(reason="Dremio only supports grants in EE/DC editions.")
```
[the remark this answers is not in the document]
bcmeireles: I can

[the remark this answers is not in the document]
howareyouman: Let's do it everywhere then

.github/scripts/create_env_file.sh
```diff
@@ -16,6 +16,9 @@ DREMIO_DATABASE=dbt_test
 DBT_TEST_USER_1=dbt_test_user_1
 DBT_TEST_USER_2=dbt_test_user_2
 DBT_TEST_USER_3=dbt_test_user_3
+DBT_TEST_ROLE_1=dbt_test_role_1
+DBT_TEST_ROLE_2=dbt_test_role_2
+DREMIO_EDITION=community
```
howareyouman, line 21: We should describe it in the docs of the tests at least. And create an enum with possible values, which may not be in this PR.

howareyouman (approved):

PR #267, opened 2025-01-21
Add devcontainers support to the repository

simonpannek:
### Description

This PR aims to introduce devcontainer support to this repository. With this addition, developers will be able to quickly set up and run a dockerized environment in VSCode, GitHub Codespaces, or any other platform that supports devcontainers.

### Changelog

-   [x] Add devcontainers support to the repository.


simonpannek:
This PR is mostly based on https://github.com/vandop/dremio-personal-analytics/ with some minor changes. We could think about kicking out Apache superset since most developers are probably not going to need it.

There should also be an easy way to replace the oss dremio container with a self-hosted enterprise version if we're working on an enterprise-only feature.

docker-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+      - DREMIO_JAVA_SERVER_EXTRA_OPTS=-Dpaths.dist=file:///opt/dremio/data/dist -Ddebug.addDefaultUser=true
+      - DREMIO_MAX_MEMORY_SIZE_MB=4096
+      - DREMIO_MAX_DIRECT_MEMORY_SIZE_MB=4096
+      - SERVICES_COORDINATOR_ENABLED=true
+      - SERVICES_EXECUTOR_ENABLED=true
+    networks:
+      - dremio-laptop-lakehouse
+    volumes:
+      - dremio_data:/opt/dremio/data:rw
+    user: "0"
```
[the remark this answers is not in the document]
bcmeireles: The user line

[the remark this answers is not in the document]
simonpannek: I mean, as long as we don't find another workaround for this work in podman, I assume we should leave it in...

99Lys (dismissed):
Great PR! I've compared with https://github.com/vandop/dremio-personal-analytics/ and saw the changes you made, they make sense to me

docker-compose.yml
```diff
@@ -0,0 +1,105 @@
+services:
```
howareyouman, line 1: Could we create a folder in the repo and move all the sh and yml files there? 
Or in leave them in .devcontainer folder?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
...
+// README at: https://github.com/devcontainers/templates/tree/main/src/python
+{
+	"name": "dbt-dremio-devcontainer",
+	// Use a Docker Compose file
+	"dockerComposeFile": "../docker-compose.yml",
+	"service": "python", 
+	"workspaceFolder": "/usr/src/app",
+
+	"runArgs": [
+  		"--userns=keep-id:uid=1000,gid=1000"
```
howareyouman, line 11: Why do we need these arguments and for what service they are?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
...
+ 	],
+ 	"containerUser": "vscode",
+	"updateRemoteUserUID": true,
+
+	// Configure tool-specific properties.
+	"customizations": {
+		"vscode": {
```
howareyouman, line 18: What if a developer doesn't have vscode installed, will it become a blocker for them to use dev containers?

docker-compose.yml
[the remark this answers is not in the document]
howareyouman: @simonpannek let's add a comment why are we doing it at least?

howareyouman (commented):
We should hide all the content for devcontainers in some folder, not in the root of the repo. It's the only requirement from my self.

Plus, have you tested it with both docker and podman?

docker-compose.yml
simonpannek: The docker-compose and setup files can be used, even if you're not using devcontainers, but I'm fine moving them into a separate folder as well. I think in that cast, I'd leave them in .devcontainer

.devcontainer/devcontainer.json
simonpannek: No, this is part of the devcontainers standard. We could also add customizations for other editors in the feature (see [here](https://containers.dev/supporting))

99Lys (approved):

bcmeireles (approved):

PR #271, opened 2025-01-28
Pass down schema on table and view creation

simonpannek:
### Description

There is no out-of-the-box way to configure dbt models materialised as tables to be created in subdirectories.

### Test Results

Added tests to check the new behavior.

### Changelog

-   [x] Push down schema definition to Dremio on table creation


simonpannek:
The test cases around this don't seem to have that great coverage. I added two basic cases, but we might want to do some manual testing/extend these before merging.

tests/utils/util.py
```diff
@@ -42,29 +41,31 @@ def relation_from_name(adapter, name: str, materialization=""):
...
+    relation_parts = name.split(".")
+
     # Make sure we have database/schema/identifier parts, even if
     # only identifier was supplied.
-    relation_parts = name.split(".")
-    if len(relation_parts) == 1:
-        if materialization == "view" or "view" in name:
-            relation_parts.insert(0, credentials.schema)
-        else:
-            relation_parts.insert(0, credentials.root_path)
```
bcmeireles, line 52: Has the logic when only the schema or root are provided been moved somewhere. Did the macro changes make this completely impossible to happen? 

Even then we should be careful about it as users could be overwriting the macro themselves and depending on this logic

tests/utils/util.py
```diff
@@ -42,29 +41,31 @@ def relation_from_name(adapter, name: str, materialization=""):
...
-        # if the relation is a view then use database
-        if materialization == "view" or "view" in name:
-            relation_parts.insert(0, credentials.database)
-            relation_type = "view"
-        else:
-            relation_parts.insert(0, credentials.datalake)
-            relation_type = "table"
+
+    # if the relation is a view then use database
+    LOGGER.info(f"Credentials: {credentials}")
```
bcmeireles, line 50: [nit]
remove credential logging

tests/utils/util.py
```diff
@@ -42,29 +41,31 @@ def relation_from_name(adapter, name: str, materialization=""):
...
         "database": relation_parts[0],
-        "schema": relation_parts[1],
-        "identifier": relation_parts[2],
+        "schema": ".".join(relation_parts[1:-1]),
+        "identifier": relation_parts[-1],
         "type": relation_type,
     }
 
+    LOGGER.info(f"Relation kwargs: {kwargs}")
+
```
bcmeireles, line 68: [nit]
remove this logging too

tests/utils/util.py
```diff
@@ -17,9 +17,8 @@
 from contextlib import contextmanager
 from dbt.tests.util import AnyInteger
 
-from dbt.adapters.events.logging import AdapterLogger
-
-logger = AdapterLogger("dremio")
+import logging
+LOGGER = logging.getLogger(__name__)
```
bcmeireles, line 21: Why have we switched from `AdapterLogger` to this?

tests/utils/util.py
simonpannek: Should've added a note on this. I altered the logging for manual testing, but I'm gonna remove it in the next commit. I'll keep this conversation open so we don't forget.

tests/utils/util.py
```diff
@@ -42,29 +41,31 @@ def relation_from_name(adapter, name: str, materialization=""):
...
+        relation_parts.insert(0, credentials.datalake)
+        relation_parts.insert(1, credentials.root_path)
+    
+    relation_type = "table" if materialization != "view" and "view" not in name else "view"
+
     kwargs = {
         "database": relation_parts[0],
-        "schema": relation_parts[1],
-        "identifier": relation_parts[2],
+        "schema": ".".join(relation_parts[1:-1]),
```
howareyouman, line 62: What if there is no 1st element? and only zero one?

howareyouman (commented):
Thank you for your PR! 

Are you planning to add more test cases or you think we are ok with only 2 you provided?

Could you please also remove info logging? If you need them, you could change logging level to  "debug".

bcmeireles (approved):

howareyouman (approved):

PR #278, opened 2025-02-24
Fix twin strategy implementation

simonpannek:
### Summary

dbt assumes tables and views share locations, unlike Dremio that keeps tables and views separately.  We address this conflict using the [twin-strategy](https://github.com/dremio/dbt-dremio/wiki/Using-Materializations-with-Dremio#user-content-optional-twin-strategy-configuration). Currently, the adapter's clone strategy seems to create a view for every table, even if there is no conflicting view to be replaced.

### Description

With this change, we introduce a check that will only overwrite existing views, not create new ones on table creation when the `clone` strategy is enabled. When a view is created with a conflicting table, the view will always select the existing table.

### Test Results

Due to this change, tests that queried the corresponding view after table creation (instead of the table directly) were failing due to this change. The schema of all affected tests have been adjusted.

### Changelog

-   [x] Only overwrite existing views in case of schema and name conflict
-   [x] Adjust tests to query created tables directly instead of their matching views


dbt/include/dremio/macros/materializations/twin_strategy.sql
```diff
@@ -24,12 +24,13 @@ limitations under the License.*/
...
-          schema=generate_schema_name_impl(target.schema, config.get('schema', validator=validation.any[basestring]), model),
-          database=generate_database_name_impl(target.database, config.get('database', validator=validation.any[basestring]), model),
-          type='view') -%}
+    {%- set view_relation = api.Relation.create(
+        identifier=generate_alias_name_impl(model.name, config.get('alias', validator=validation.any[basestring]), model),
+        schema=generate_schema_name_impl(target.schema, config.get('schema', validator=validation.any[basestring]), model),
+        database=generate_database_name_impl(target.database, config.get('database', validator=validation.any[basestring]), model),
+        type='view') -%}
+    {%- set conflicting_view = adapter.get_relation(database=view_relation.database, schema=view_relation.schema, identifier=view_relation.identifier) -%}
+    {%- if conflicting_view is not none -%}
```
simonpannek, line 33: This is the main change of this PR. All other adjustments are test fixes since quite a few tests target the matching view after creating a table. Since we're not creating a view anymore, these tests will break.

tests/functional/adapter/basic/test_data_types.py
```diff
@@ -12,9 +12,36 @@
...
+
+        # check contents equal
+        check_relations_equal(project.adapter, ["expected", "actual_view"])
+
+        # check types equal
+        expected_cols = get_relation_columns(project.adapter, "expected")
+        actual_cols = get_relation_columns(project.adapter, "actual_view")
+        print(f"Expected columns: {expected_cols}")
+        print(f"Actual columns: {actual_cols}")
+        self.assert_columns_equal(project, expected_cols, actual_cols)
```
simonpannek, line 47: This is just overwriting the same function from the dbt testing code to use our custom `get_relation_columns` implementation instead of the dbt one.

dbt/include/dremio/macros/materializations/twin_strategy.sql
```diff
@@ -14,25 +14,45 @@ limitations under the License.*/
...
-          type='table') -%}
-      {{ adapter.drop_relation(table_relation) }}
+    -- Check if there is a conflicting table
+    {%- set table_relation = api.Relation.create(
+        identifier=generate_alias_name_impl(model.name, config.get('file', validator=validation.any[basestring]), model),
+        schema=generate_schema_name_impl(target.root_path, config.get('root_path', validator=validation.any[basestring]), model),
+        database=generate_database_name_impl(target.datalake, config.get('datalake', validator=validation.any[basestring]), model),
+        type='table') -%}
+    {%- set conflicting_table = adapter.get_relation(database=table_relation.database, schema=table_relation.schema, identifier=table_relation.identifier) -%}
+    {%- if conflicting_table is not none -%}
```
howareyouman, line 32: Could we reduce if\else statements and use only 1 level of integrity?

tests/functional/adapter/basic/test_base_mat.py
```diff
@@ -36,7 +36,7 @@
 version: 2
 sources:
   - name: raw
-    database: "dbt_test_source"
+    database: "{{ target.datalake }}"
```
howareyouman, line 39: Finally!

tests/functional/adapter/basic/test_docs_generate.py
```diff
@@ -43,6 +48,22 @@
 select * from {{ ref('seed') }}
 """
 
+ref_sources__schema_yml = """
+version: 2
+sources:
+  - name: my_source
+    description: "{{ doc('source_info') }}"
+    loader: a_loader
```
howareyouman, line 56: For the future reviewers - loader is just a description of EL part of ELT. Description only.

tests/utils/util.py
```diff
@@ -79,6 +82,14 @@ def get_connection(adapter, name="_test"):
         conn = adapter.connections.get_thread_connection()
         yield conn
 
+# Uses:
```
howareyouman, line 85: Why do we need this documentation?

howareyouman (approved):
Awesome tests and the fix itself! Great job @simonpannek !

howareyouman:
The last missing part here may be - what should be the safest strategy for customers to adopt this bug in the newest version? We could write a page here about it, or in the public issues.

tests/utils/util.py
simonpannek: This is just copied from the `util.py` dbt ships. Should I remove it?

howareyouman: Let's remove it. It seems pretty useless.

bcmeireles (dismissed):

tests/functional/adapter/basic/test_concurrency.py
```diff
@@ -21,13 +21,13 @@
...
 )
 from dbt.tests.util import (
-    check_relations_equal,
     check_table_does_not_exist,
     rm_file,
     run_dbt,
     run_dbt_and_capture,
     write_file,
 )
+from tests.utils.util import check_relations_equal
```
99Lys, line 30: What is the difference between `dbt.tests.util` and `tests.utils.util`?

99Lys (dismissed):

tests/functional/adapter/basic/test_concurrency.py
simonpannek: `check_relations_equal` in `dbt.tests.util` uses our custom `relation_from_name` implementation, not the one shipping with dbt. Adding a comment for clarification

dbt/include/dremio/macros/materializations/twin_strategy.sql
simonpannek: Done

tests/utils/util.py
simonpannek: Done

howareyouman (dismissed):

howareyouman (dismissed):
I don't see any other major changes from the last review.

simonpannek:
@howareyouman after rebasing, we had new failing tests, so these had to be adjusted. I also just realized we released 1.8.2 in the meanwhile, so this change needs to be moved to 1.8.3

howareyouman (approved):

bcmeireles (approved):

PR #281, opened 2025-03-12
dbt Constraints / model contracts

bcmeireles:
### Summary

Related to https://github.com/dbt-labs/dbt-core/issues/6079, allows defining column types and constraints on columns in Dremio, supporting the dbt features of model contracts and constraints

### Description

- Added support to enforce column names and data types
- Retrieves data types and constraints defined in the model YML file and use it to generate the model DDL
- Dremio only supports the `not null` constraint and it is currently not enforced

### Test Results

Added a total of 18 new tests, inherited from dbt-core. 2 of these (rollback) are currently being skipped as Dremio does not enforce any constraints at the moment

### Changelog


dbt/include/dremio/macros/materializations/table/create_table_as.sql
```diff
@@ -41,6 +39,13 @@ limitations under the License.*/
   {{ config_cols("localsort by") }}
   {{ store_as_clause() }}
   {{ single_writer_clause() }}
+
+  {% if contract_config.enforced %}
+     {{ get_table_columns_and_constraints() }}
+     {{ get_assert_columns_equivalent(sql) }}
+     {% set sql = get_select_subquery(sql) %}
+  {% endif %}
```
bcmeireles, line 47: `get_select_subquery` is called here so that the order the columns are defined in the yaml does not need to be the same as in the sql query. This isn't mentioned in the [Maintainers Guide to Upgrade to 1.5.0](https://github.com/dbt-labs/dbt-core/discussions/7213) as this enhancement was added later

dbt/include/dremio/macros/materializations/view/create_view_as.sql
```diff
@@ -14,14 +14,15 @@ limitations under the License.*/
...
   {%- set sql_header = config.get('sql_header', none) -%}
 
   {{ sql_header if sql_header is not none }}
 
-  create or replace view {{ relation }} as {{ sql }}
+  create or replace view {{ relation }}
+  {% if contract_config.enforced %}
+     {{ get_assert_columns_equivalent(sql) }}
+     {% set sql = get_select_subquery(sql) %}
+  {% endif %}
```
bcmeireles, line 25: Same as the comment above

dbt/adapters/dremio/impl.py
```diff
@@ -42,6 +44,14 @@ class DremioAdapter(SQLAdapter):
     Relation = DremioRelation
     Column = DremioColumn
 
+    CONSTRAINT_SUPPORT = {
+        ConstraintType.check: ConstraintSupport.NOT_SUPPORTED,
+        ConstraintType.not_null: ConstraintSupport.NOT_ENFORCED,
+        ConstraintType.unique: ConstraintSupport.NOT_SUPPORTED,
+        ConstraintType.primary_key: ConstraintSupport.NOT_SUPPORTED,
+        ConstraintType.foreign_key: ConstraintSupport.NOT_SUPPORTED,
+    }
```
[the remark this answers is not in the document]
bcmeireles: Currently when creating tables Dremio only supports defining the `not null` constraint even though it does not enforce it. Once that behavior is present in Dremio we should update this to `ConstraintType.not_null: ConstraintSupport.ENFORCED,`

tests/functional/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,152 @@
...
+class TestDremioTableConstraintsRuntimeDdlEnforcement(
+        DremioColumnEqualSetup, BaseConstraintsRuntimeDdlEnforcement):
+    @pytest.fixture(scope="class")
+    def expected_sql(self):
+        return _expected_sql_dremio
+
+
+# TODO: Revisit once at least one of the constraints starts being enforced
+@pytest.mark.skip("Dremio does not enforce any constraints so rollbacks can't be tested")
+class TestDremioTableConstraintsRollback(BaseConstraintsRollback):
```
bcmeireles, line 98: Once Dremio starts enforcing the `not null` constraint we can stop skipping these tests, I decided to leave them here even though skipped to future-proof this as they could be easy to miss later on

tests/functional/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,152 @@
...
+
+class TestDremioIncrementalConstraintsColumnsEqual(
+        DremioColumnEqualSetup,
+        BaseIncrementalConstraintsColumnsEqual):
+    pass
+
+
+# TODO: Revisit once at least one of the constraints starts being enforced
+@pytest.mark.skip("Dremio does not enforce any constraints so rollbacks can't be tested")
+class TestDremioIncrementalConstraintsRollback(BaseIncrementalConstraintsRollback):
```
bcmeireles, line 110: Same reason as above for this skip

tests/functional/adapter/constraints/test_constraints.py
99Lys, line 1: Could you please explain how the tests are working here? Are we performing any tests here (or they are on dbt-core's side)? Asking because it's all `return`s and `pass`. Thanks!!

bcmeireles: They are inherited from dbt core itself. The tests that run are [these](https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/adapter/constraints/test_constraints.py). Using `pass` just means that you're running them exactly as inherited. The `return`s aren't in the actual tests but on the fixtures, to override them, ie `expected_sql` as Dremio has a different SQL from the base dbt one, which overrides [this one](https://github.com/dbt-labs/dbt-core/blob/9d7820c356e6bbdd8af756df1fc1b1a2c086b857/tests/functional/adapter/constraints/test_constraints.py#L200-L227). This is how all the other dbt tests work. By inheriting I mean cases like `class TestDremioTableConstraintsRollback(BaseConstraintsRollback)` that mean that the `TestDremioTableConstraintsRollback` test suite will run the tests inside dbt's `BaseConstraintsRollback`, which can be found [here](https://github.com/dbt-labs/dbt-core/blob/9d7820c356e6bbdd8af756df1fc1b1a2c086b857/tests/functional/adapter/constraints/test_constraints.py#L253)

tests/functional/adapter/constraints/test_constraints.py
simonpannek: What is our take on skipping tests vs. marking them as explicitly failing right now?

bcmeireles: For functionality that Dremio does not support we have examples where we are skipping:
https://github.com/dremio/dbt-dremio/blob/16ccee374c1fbea54ca3c10dab8096bf79ba7992/tests/functional/adapter/utils/test_array_construct.py#L20-L22

Also my idea is that we're actively handling and decreasing the expected failures so we can move away from such logic in the pipeline. I'd say we should not add to the expected failures and that it would either be this or not have the test at all (since unsupported, but by removing this it's easier to miss adding it back once Dremio starts enforcing `not null`)

99Lys (approved):

simonpannek (approved):

tests/functional/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,152 @@
...
+            ["1", schema_int_type, int_type],
+            ["'1'", string_type, string_type],
+            ["cast('2019-01-01' as date)", "date", "DATE"],
+            ["true", "boolean", "BOOLEAN"],
+            ["cast('2013-11-03 00:00:00-07' as TIMESTAMP)", "timestamp(6)", "TIMESTAMP"],
+            ["cast('1' as DECIMAL)", "DECIMAL", "DECIMAL"],
+        ]
+
+    @pytest.fixture(scope="class")
+    def dbt_profile_data(
```
howareyouman, line 56: Am I right that this function will no longer be needed after Simon's subfolder feature merged?

tests/functional/adapter/constraints/test_constraints.py
```diff
@@ -0,0 +1,152 @@
...
+        select 
+            'blue' as color, 
+            1 as id, 
+            '2019-01-01' as date_day 
+        ) as model_subq 
+    )
+"""
+
+
+class DremioColumnEqualSetup:
```
howareyouman, line 34: Could you please add documentation here about this class, I don't understand why do we need it.

tests/functional/adapter/constraints/test_constraints.py
howareyouman: In general, I see our strategy as - remove skips as more as possible and let the tests fail. 
The skips are fine if we have at least P2 ticket with them and we know how to fix them.

howareyouman (approved):
Could you please add more documentation in the test, but overall - great work, thanks!

howareyouman:
@bcmeireles Could you please update the changelog as well?

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: I'll create a ticket for these and move it to backlog. Do note that this ticket will be blocked until dremio starts enforcing `not null`

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: I'll be rebasing so we can confirm

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: All the tests required these fixture overrides so instead of pasting the code in all of them I made a class out of them, that then gets inherited by the tests to avoid repetition. I'll add a comment explaining this

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: I'm in favor of having expected fails if it's an issue with the test itself that must be fixed, in this case it's a test with something that Dremio does not currently support so I think skipping in these cases would be best. 

DX-101860 was created

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: I've just tried this. This fixture is still needed with the latest changes

simonpannek: Yeah, it's really weird - so we do still need it, but sometimes it's fine to import it from our `utils.py` and sometimes we need to explicitly overwrite it. I haven't done the investigation yet on why this is the case

simonpannek (approved):

tests/functional/adapter/constraints/test_constraints.py
howareyouman: Let's create a ticket for it then.

howareyouman (approved):

tests/functional/adapter/constraints/test_constraints.py
bcmeireles: DX-101866

PR #282, opened 2025-03-13
Fixed column schemas not being grabbed properly

### LINKED ISSUE #20: TestSnapshotCheckColsDremio Failing
This test is failing after merging with the rest-api changes. Currently getting an error parsing the following query, which creates a staging table for snapshots:

  ```
create table "rav-test"."rav-test"."test16656264419696928231_test_basic"."cc_all_snapshot__dbt_tmp"
  as (
    with snapshot_query as (    
    select * from "rav-test"."rav-test"."test16656264419696928231_test_basic"."added"
    ),
    snapshotted_data as (
        select *,
            id as dbt_unique_key
        from "rav-test"."rav-test"."test16656264419696928231_test_basic"."cc_all_snapshot"
        where dbt_valid_to is null
    ),
    insertions_source_data as (
        select
            *,
            id as dbt_unique_key,
            (SELECT CURRENT_TIMESTAMP()) as dbt_updated_at,
            (SELECT CURRENT_TIMESTAMP()) as dbt_valid_from,
            nullif((SELECT CURRENT_TIMESTAMP()), (SELECT CURRENT_TIMESTAMP())) as dbt_valid_to,
            md5(coalesce(cast(id as varchar ), '')
         || '|' || coalesce(cast((SELECT CURRENT_TIMESTAMP()) as varchar ), '')
        ) as dbt_scd_id
        from snapshot_query
    ),
    updates_source_data as (
        select
            *,
            id as dbt_unique_key,
            (SELECT CURRENT_TIMESTAMP()) as dbt_updated_at,
            (SELECT CURRENT_TIMESTAMP()) as dbt_valid_from,
            (SELECT CURRENT_TIMESTAMP()) as dbt_valid_to

        from snapshot_query
    ),
    insertions as (
        select
            'insert' as dbt_change_type,
            source_data.*
        from insertions_source_data as source_data
        left outer join snapshotted_data on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where snapshotted_data.dbt_unique_key is null
           or (
                snapshotted_data.dbt_unique_key is not null
            and (
                ()
            )
        )

    ),
    updates as (
        select
            'update' as dbt_change_type,
            source_data.*,
            snapshotted_data.dbt_scd_id

        from updates_source_data as source_data
        join snapshotted_data on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where (
            ()
        )
    )
    select * from insertions
    union all
    select * from updates
  )
```

The parse error occurs at `and ( () )` under "insertions as" and `where( () )` under "updates as". In place of these empty parentheses should be something similar to `snapshotted_data.some_date != source_data.some_date`. This value is populated by a macro created by dbt-core. The exact value is called strategy.rows_changed. It is supposed to contain a list of columns where the row has changed, but my guess is our cursor object is not correctly getting all the columns.

### PR DESCRIPTION
**Pull request by bcmeireles**
### Summary

Fixes the column schema not being properly grabbed from query results

### Description

Previously dbt was not able to get the schema of the columns interacted with in a query, these changes fix that and we are now able to get information about the columns (names and data types)

### Test Results

3 expected failures fixed
2 skipped tests that were also failing are not skipped anymore

### Changelog

### Contributor License Agreement

-   [x] Please make sure you have signed our [Contributor License Agreement](https://www.dremio.com/legal/contributor-agreement/), which enables Dremio to distribute your contribution without restriction.

### Related Issue

Fixes #20 and also some other tests

### CODE REVIEW

**Review by simonpannek — DISMISSED** [review]

File: CHANGELOG.md [hunk 1]
```diff
@@ -8,10 +8,12 @@
 - Adds `BaseIncrementalOnSchemaChange` test to test_incremental.py
 - Changed logic for partitioning when materializing tables. Double quoting issue has been removed, now letting the user decide the quoting
     - New example: `partition_by=['month("datetime_utc")']`
+- Fixed grabbing column schemas
```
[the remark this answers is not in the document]
bcmeireles: Added some more context directly on the changelog. This is mostly invisible to users/requires no action on their part

simonpannek (approved):

dbt/adapters/dremio/column.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+from dbt.adapters.base.column import Column
+
+@dataclass(init=False)
+class DremioColumn(Column):
+    TYPE_LABELS = {
+        "TEXT": "VARCHAR",
```
howareyouman, line 8: The only problem I have here is - will it block other types or not. 
Could you please add more references to other connectors?

dbt/adapters/dremio/impl.py
```diff
@@ -157,6 +160,17 @@ def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
                 grants_dict.update({privilege: [f"{grantee_type}:{grantee}"]})
         return grants_dict
 
+    def get_column_schema_from_query(self, sql: str) -> List[BaseColumn]:
+        """Get a list of the Columns with names and data types from the given sql."""
+        _, cursor = self.connections.add_select_query(sql)
+        columns = [
+            self.Column.create(
+                entry['name'], self.connections.data_type_code_to_name(entry['type']['name'])
```
howareyouman, line 168: What if `entry['name']` or `entry['type']['name']` are missing?

dbt/adapters/dremio/impl.py
```diff
@@ -157,6 +160,17 @@ def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
...
+            )
+            for entry in cursor.job_results()['schema']
```
howareyouman, line 170: Is `cursor.job_results()['schema']` a mandatory field as well?

howareyouman (commented):
In general I have no blockers from my side, but it would be awesome to check all the possible problems we may introduce with it.

dbt/adapters/dremio/column.py
bcmeireles: This is kind of just a translation in case the dbt-assumed syntax is not valid for the data platform. [bigquery](https://github.com/dbt-labs/dbt-bigquery/blob/0995665e490cdee9c408d26aac8e1c19fefaebe0/dbt/adapters/bigquery/column.py#L15-L20) only has 3 of these, [databricks](https://github.com/databricks/dbt-databricks/blob/7c68e99563553b70ecfeb5dc9cbf0307299d7beb/dbt/adapters/databricks/column.py#L13-L15) only had to do it for one but other adapters such as [snowflake](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/adapters/snowflake/column.py) didn't even need any translations. By running all the available tests we only had problems with `TEXT` and `STRING`. which had to be translated into `VARCHAR`

dbt/adapters/dremio/impl.py
bcmeireles: These will always come populated as per adapter behavior. The dbt function we override also doesn't do any extra checks in here because if any of these came empty then it'd be because of no data regarding the columns, we'd have to return an empty list and it'd fail later because it'd be expecting it. Per adapter behavior, if we reach this function, we have the fields to read

dbt/adapters/dremio/impl.py
bcmeireles: What do you mean by mandatory? If we reach this function, we have `schema` in job results populated

howareyouman (approved):

PR #285, opened 2025-03-14
Add strategies to reflections

bcmeireles:
### Summary

Currently when creating a reflection, dbt marks the job as done as soon as it sends the API call to Dremio. This new option allows users to only continue after the reflection is fully created on Dremio's side.

### Description

- Introduces a new strategy: `reflection_strategy` with 3 valid values: `trigger`, `wait` and `depend`. 
  - `trigger` is the default and keeps the old behavior
  - `wait` will try to wait for the reflection to be live, up to a maximum of `max_wait_time` seconds, `30` being the default value
  - `depend` will not proceed until the reflection is live
- New config options for reflections:
  - `max_wait_time`: Maximum time in seconds that the `wait` strategy will wait for the reflection to be ready. Default value is `30`
  - `check_interval`: Time between the API calls that check if a reflection is ready. Default value is `5`

### Test Results

Added 2 extra tests, one for `trigger` and another one for `wait`

### Changelog

### Related Issue

Suggested in #184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+# dbt-dremio v1.8.3
+
+## Changes
+
+- Added 3 strategies for reflections: `trigger`, `wait` and `depend`
+  - `trigger` keeps the previous behavior for reflections, where the job to create them is triggered and then dbt-dremio moves on to other models
+  - `wait` will wait for the reflection to be created before moving on to other models, up to a `max_wait_time` timeout
+  - `depend` will wait for the reflection to be created before moving on to other models, with no timeout
+  - New fields in the reflections config:
+    - `strategy`: `trigger`, `wait` or `depend`
+    - `max_wait_time`: time in seconds to wait for the reflection to be created before moving on to other models
+      - default: `30`
+    - `wait_interval`: time in seconds to wait between checks for the reflection to be created
+      - default: `5`
+
+## Features
+
+- [#285](https://github.com/dremio/dbt-dremio/pull/285) Add strategies to reflections
+
 # dbt-dremio v1.8.2
 
 ## Changes
--- a/tests/functional/adapter/dremio_specific/test_reflections.py
+++ b/tests/functional/adapter/dremio_specific/test_reflections.py
@@ -434,6 +434,8 @@ class TestReflectionsDremio:
         (results, log_output) = run_dbt_and_capture(["run", "--select", "view1", "wait_strategy_timeout_reflection"])
         assert "did not become available within 1 seconds, skipping wait" in log_output
 
+    # This test will check if the adapter does not wait for the reflection to become available when using the trigger strategy
+    # If we waited, the log line being checked would be present
     def testTriggerStrategyTimeoutReflection(self, project):
         (results, log_output) = run_dbt_and_capture(["run", "--select", "view1", "trigger_strategy_timeout_reflection"])
         assert "did not become available within 1 seconds, skipping wait" not in log_output


tests/functional/adapter/dremio_specific/test_reflections.py
```diff
@@ -407,3 +429,11 @@ def testNameReflectionFromFilename(self, project, client):
         assert "partitionFields" not in reflection
         assert "sortFields" not in reflection
         assert reflection["partitionDistributionStrategy"] == "STRIPED"
+
+    def testWaitStrategyTimeoutReflection(self, project):
+        (results, log_output) = run_dbt_and_capture(["run", "--select", "view1", "wait_strategy_timeout_reflection"])
+        assert "did not become available within 1 seconds, skipping wait" in log_output
+
+    def testTriggerStrategyTimeoutReflection(self, project):
```
99Lys, line 437: Are we checking `max_wait_time` when strategy is `trigger`? Because in `create_reflection` method in `connections.py` there are only 2 conditions - for wait and depend respectively

bcmeireles: The `trigger` strategy does not wait for the reflection to be materialized before continuing so there is nothing to wait for in that case

99Lys: In that case could you please adjust this test for this distinction to be more clear? Or add some comments explaining it because it seems like we are checking `max_wait_time` in `trigger`

bcmeireles: Added a comment to the test

bcmeireles: Added a comment to the test, we are in fact checking for `max_wait_time` but we are checking that it was **not** used

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+
+- Added 3 strategies for reflections: `trigger`, `wait` and `depend`
+  - `trigger` keeps the previous behavior for reflections, where the job to create them is triggered and then dbt-dremio moves on to other models
+  - `wait` will wait for the reflection to be created before moving on to other models, up to a `max_wait_time` timeout
+  - `depend` will wait for the reflection to be created before moving on to other models, with no timeout
+  - New fields in the reflections config:
+    - `strategy`: `trigger`, `wait` or `depend`
+    - `max_wait_time`: time in seconds to wait for the reflection to be created before moving on to other models
+      - default: `30`
+    - `wait_interval`: time in seconds to wait between checks for the reflection to be created
```
simonpannek, line 13: I think you updates this to be called `check_interval` instead

dbt/adapters/dremio/connections.py
```diff
@@ -326,7 +326,7 @@ def create_reflection(self, name: str, reflection_type: str, anchor: DremioRelat
                           date_dimensions: List[str], measures: List[str],
                           computations: List[str], partition_by: List[str], partition_transform: List[str],
                           partition_method: str, distribute_by: List[str], localsort_by: List[str],
-                          arrow_cache: bool) -> None:
+                          arrow_cache: bool, reflection_strategy: str, max_wait_time: int, check_interval: int) -> None:
```
simonpannek, line 329: Should the connector check for constrains of the values provided here? For instance, a negative or 0 value of `check_interval` might lead to errors or accidental rate limiting by Dremio. (See DC rate limits [here](https://docs.dremio.com/cloud/help-support/appendix/limits/#rate-limits))

tests/functional/adapter/dremio_specific/test_reflections.py
```diff
@@ -170,6 +190,8 @@ def models(self):
             "default_displays_model.sql": default_displays_model,
             "name_reflection_from_alias.sql": name_reflection_from_alias_model,
             "name_reflection_from_filename.sql": name_reflection_from_filename_model,
+            "wait_strategy_timeout_reflection.sql": wait_strategy_timeout_reflection,
+            "trigger_strategy_timeout_reflection.sql": trigger_strategy_timeout_reflection,
```
simonpannek, line 194: I realize testing `depend` is not that easy, but there must be a way. Is it possible to create a reflection that will never be finish materializing and then try to depend on it?

tests/functional/adapter/dremio_specific/test_reflections.py
```diff
@@ -407,3 +429,13 @@ def testNameReflectionFromFilename(self, project, client):
         assert "partitionFields" not in reflection
         assert "sortFields" not in reflection
         assert reflection["partitionDistributionStrategy"] == "STRIPED"
+
+    def testWaitStrategyTimeoutReflection(self, project):
+        (results, log_output) = run_dbt_and_capture(["run", "--select", "view1", "wait_strategy_timeout_reflection"])
+        assert "did not become available within 1 seconds, skipping wait" in log_output
```
simonpannek, line 435: Can we get a positive test case for the wait strategy as well?

dbt/adapters/dremio/connections.py
bcmeireles: Will be adding a `> 0` constraint to this

tests/functional/adapter/dremio_specific/test_reflections.py
bcmeireles: The idea of `depend` is that it will be stuck at the job running that reflection until it is complete. Even if we could make a reflection that would never materialize we'd just be stuck waiting for it forever

tests/functional/adapter/dremio_specific/test_reflections.py
bcmeireles: We could but I was avoiding adding it as a reflection going through is already tested by all the other reflection tests + it has the potential to become flaky if it takes longer than expected to be materialized

CHANGELOG.md
bcmeireles: Good call

tests/functional/adapter/dremio_specific/test_reflections.py
simonpannek: My idea would've been to assert that the function hasn't returned after a few seconds and then killing the thread. But I don't know how complex we want this to get

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+# dbt-dremio v1.8.3
+
+## Changes
+
+- Added 3 strategies for reflections: `trigger`, `wait` and `depend`
+  - `trigger` keeps the previous behavior for reflections, where the job to create them is triggered and then dbt-dremio moves on to other models
+  - `wait` will wait for the reflection to be created before moving on to other models, up to a `max_wait_time` timeout
```
howareyouman, line 7: Could you please add here what will happen after `max_wait_time` is over.

dbt/adapters/dremio/connections.py
howareyouman, line 329: Should we define the default values here or in the template? What do you think?

dbt/adapters/dremio/connections.py
```diff
@@ -357,13 +357,41 @@ def create_reflection(self, name: str, reflection_type: str, anchor: DremioRelat
...
+                created_reflection = rest_client.update_reflection(reflection.get("id"), payload)
                 updated = True
                 break
 
         if not updated:
             logger.debug(f"Reflection {name} does not exist. Creating it")
-            rest_client.create_reflection(payload)
+            created_reflection = rest_client.create_reflection(payload)
+
+        if reflection_strategy == "wait":
```
howareyouman, line 368: Let's use match pattern here. And put all possible values somewhere in enum

Updated. I saw a validation in the Jinja, let's add match here at least. If/elif seems like a bad design pattern

dbt/adapters/dremio/connections.py
```diff
@@ -357,13 +357,41 @@ def create_reflection(self, name: str, reflection_type: str, anchor: DremioRelat
...
+                    return
+
+                time.sleep(check_interval)
+
+            logger.info(f"Reflection {name} did not become available within {max_wait_time} seconds, skipping wait")
+
+        elif reflection_strategy == "depend":
+            reflection_id = created_reflection["id"]
+
+            while True:
```
howareyouman, line 387: Should we in this case give some state of the reflection - how much time left? What is the status? Is everything ok?
Because while true without any status update will be painful for customers.

tests/functional/adapter/dremio_specific/test_reflections.py
howareyouman: Yes, I agree with Simon that we should have guarantees that it will end. Waiting without a proof for a green result may be frustrating for customers.

tests/functional/adapter/dremio_specific/test_reflections.py
howareyouman: Without a proper mocks these tests could be all eventually flaky

howareyouman (commented):
We should polish depend strategy somehow before the release (or at least create a ticket for it)

tests/functional/adapter/dremio_specific/test_reflections.py
bcmeireles: That would be the `wait` strategy where the customer would set the maximum time they'd want to wait. Should we remove `depend` in that case? Or have it act as `wait` but throw an error instead of just skipping?

dbt/adapters/dremio/connections.py
bcmeireles: They are being defined in the `reflection` materialization, `reflection.sql` has the following lines:
```py
 {%- set max_wait_time = config.get('max_wait_time', validator=validation.any[int]) or 30 -%}
 {%- set check_interval = config.get('check_interval', validator=validation.any[int]) or 5 -%}
```
This is on pair with how we do it in other situations

dbt/adapters/dremio/connections.py
bcmeireles: Not exactly sure if I'm seeing what you're suggesting here?

dbt/adapters/dremio/connections.py
bcmeireles: Maybe we could show the `status` -> `combinedStatus` that is given by the API response to our call. Docs [here](https://docs.dremio.com/current/reference/api/reflections/)

howareyouman: We could ask the question to our reflections team about it

bcmeireles: I've reached out to them 👍

PR #286, opened 2025-03-17
Adding new check if the space already exists

howareyouman:
### Summary

Adding check if catalog/space already exists. It will unblock non-admin users in DC from starting to use dbt-dremio. 

### Description

Before only ADMIN users of DC were able to run dbt-dremio. Now we are checking if the catalog exists and it unblocks customers to use non-admin users in DC with dbt-dremio.

### Test Results

Created a tech dept ticket to create dbt-dremio tests for it. Run only manual tests with DC instance.

### Changelog

### DIFF
--- a/dbt/adapters/dremio/connections.py
+++ b/dbt/adapters/dremio/connections.py
@@ -224,8 +224,8 @@ class DremioConnectionManager(SQLConnectionManager):
         database = relation.database
         schema = relation.schema
 
-        if database == ("@" + credentials.UID) or self.__catalog_exists(
-            [database]):
+        if database == ("@" + credentials.UID) or self._catalog_exists(
+            database):
             logger.debug(
                 "Database is default or already exists: creating folders only")
         else:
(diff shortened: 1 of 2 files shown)


simonpannek (approved):

dbt/adapters/dremio/connections.py
```diff
@@ -224,8 +224,10 @@ def create_catalog(self, relation):
         database = relation.database
         schema = relation.schema
 
-        if database == ("@" + credentials.UID):
-            logger.debug("Database is default: creating folders only")
+        if database == ("@" + credentials.UID) or self.__catalog_exists(
+            [database]):
```
bcmeireles, line 228: Would it be better to have the `__catalog_exists` function check if we are passing a string instead of a list as for the path? Feel like it would be more extendable to convert a single string into the expected list in the function itself in case we have more paths structured like this in the future, instead of converting into a list in the function call itself

dbt/adapters/dremio/connections.py
```diff
@@ -234,7 +236,20 @@ def create_catalog(self, relation):
             logger.debug(f"Creating folder(s): {database}.{schema}")
             self._create_folders(database, schema, rest_client)
         return
-    
+
+    def __catalog_exists(self, path):
+        thread_connection = self.get_thread_connection()
```
bcmeireles, line 241: Any specific reason for the `__` instead of `_` like we have for other functions?

bcmeireles (commented):
Guess merging without tests will be OK for now as we don't have the DC pipelines but we should keep in mind to come back to this once we have them

dbt/adapters/dremio/connections.py
howareyouman: That's a good question. I was curious about codestyle as well. Let's create a ticket to add checkstyle validation to dbt-dremio.

howareyouman: Created ticket for you @bcmeireles DX-101815

bcmeireles (approved):

99Lys (approved):

PR #289, opened 2025-04-02
Include dbt-dremio version in query metadata

99Lys:
### Summary

We are trying to understand the adoption and distribution of dbt-dremio from DC usage data, therefore, we are including an additional value in query metadata that represents the version of dbt-dremio adapter.

Metadata before changes: `/* {"app": "dbt", "dbt_version": "1.8.8", "profile_name": <profile_name>, "target_name": <target_name>, ... } */`

Metadata after changes: `/* {"app": "dbt", "dbt_version": "1.8.8", "dbt_dremio_version": "1.8.3", "profile_name": <profile_name>, "target_name": <target_name>, ... } */`

### Description

- Created a new macro `DREMIO_QUERY_COMMENT`, which includes the dbt-dremio's version, to replace the default query comment `DEFAULT_QUERY_COMMENT` macro
- Created a new class `DremioMacroQueryStringSetter` that inherits `MacroQueryStringSetter` to override the method that configures the query comment macro to be evaluated

### Test Results

All tests are passing, no new tests added - query comment tests already exists and no updates are needed

### Changelog

### DIFF
--- a/dbt/adapters/dremio/connections.py
+++ b/dbt/adapters/dremio/connections.py
@@ -16,7 +16,9 @@ import agate
 from typing import Any, Dict, Tuple, Optional, List
 from contextlib import contextmanager
 
-from dbt.adapters.dremio import __version__
+from dbt.adapters.base.query_headers import MacroQueryStringSetter
+
+from dbt.adapters.dremio.__version__ import version
 from dbt.adapters.dremio.api.cursor import DremioCursor
 from dbt.adapters.dremio.api.handle import DremioHandle
 from dbt.adapters.dremio.api.parameters import ParametersBuilder
@@ -30,7 +32,7 @@ import json
 
 import dbt_common.exceptions
 from dbt.adapters.sql import SQLConnectionManager
-from dbt.adapters.contracts.connection import AdapterResponse
+from dbt.adapters.contracts.connection import AdapterResponse, DEFAULT_QUERY_COMMENT
 
 from dbt.adapters.dremio.api.rest.client import DremioRestClient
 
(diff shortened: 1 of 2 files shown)


bcmeireles (approved):

simonpannek (approved):
Did you do manual testing for this? Have you confirmed this works both for EE and DC?

dbt/adapters/dremio/connections.py
```diff
@@ -48,12 +51,43 @@
...
+  {{# in the node context, the connection name is the node_id #}}
+  {{%- do comment_dict.update(connection_name=connection_name) -%}}
+{{%- endif -%}}
+{{{{ return(tojson(comment_dict)) }}}}
+"""
+
+class DremioMacroQueryStringSetter(MacroQueryStringSetter):
+    # Overriding this method to update the query comment macro
+    def _get_comment_macro(self) -> Optional[str]:
+        if self.config.query_comment.comment == DEFAULT_QUERY_COMMENT:
```
howareyouman, line 77: Why do we need this if statement? Should we always use custom comment?

howareyouman (commented):
It would be awesome to add at least 1 test for this function.

dbt/adapters/dremio/connections.py
99Lys: I suppose that since we've now defined a custom macro for this, it would replace the previous default one, so we will use `DREMIO_QUERY_COMMENT` as default from now.
The if statement exists to prevent our custom macro to override query comments configured by a user → check https://docs.getdbt.com/reference/project-configs/query-comment

99Lys:
> Did you do manual testing for this? Have you confirmed this works both for EE and DC?

- Ran pipeline against EE and manually ran DC, check ticket for details

---

> It would be awesome to add at least 1 test for this function.

We already had tests for this in our adapter → `tests/functional/adapter/query_comment/test_query_comment.py`
Each test handles different types of query comment macros and all are passing → check `dbt-adapters/dbt-tests-adapter/src/dbt/tests/adapter/query_comment/fixtures.py`

PR #291, opened 2025-04-21
dbt Unit Testing - basic implementation

99Lys:
### Summary

Introducing unit testing feature

### Description

- Adjusted standard implementation to allow unit testing against models materialized as views
- With this current implementation, unit testing feature still has limitations:
    - Only **tables** and **views** without dependencies, i.e., without using Jinja function `ref()`, can be tested
    - Not all data types / castings are supported

### Test Results

- All tests are passing, adjusted existing unit tests to test the ability to run unit tests against models without dependencies

### Changelog


.github/expected_failures.txt
```diff
@@ -2,9 +2,7 @@ tests/component/test_profile_template.py::TestProfileTemplate::test_cloud_option
 tests/functional/adapter/dremio_specific/test_drop_temp_table.py::TestDropTempTableDremio::test_drop_temp_table
 tests/functional/adapter/dremio_specific/test_schema_parsing.py::TestSchemaParsingDremio::test_schema_with_dots
 tests/functional/adapter/dremio_specific/test_verify_ssl.py::TestVerifyCertificateDremio::test_insecure_request_warning_not_exist
-tests/functional/adapter/unit_testing/test_unit_testing.py::TestDremioUnitTestingTypes::test_unit_test_data_type
 tests/functional/adapter/unit_testing/test_unit_testing.py::TestDremioUnitTestCaseInsensitivity::test_case_insensitivity
-tests/functional/adapter/unit_testing/test_unit_testing.py::TestDremioUnitTestInvalidInput::test_invalid_input
```
bcmeireles, line 7: 2 more tests done! This is great :)

tests/functional/adapter/unit_testing/test_unit_testing.py
```diff
@@ -19,65 +19,206 @@
...
+            # ],
             # array of struct
-            [
-                "[struct(st_geogpoint(75, 45) as my_point), struct(st_geogpoint(75, 35) as my_point)]",
-                "['struct(st_geogpoint(75, 45) as my_point)', 'struct(st_geogpoint(75, 35) as my_point)']",
-            ],
+            # [
+            #     "[struct(st_geogpoint(75, 45) as my_point), struct(st_geogpoint(75, 35) as my_point)]",
+            #     "['struct(st_geogpoint(75, 45) as my_point)', 'struct(st_geogpoint(75, 35) as my_point)']",
+            # ],
```
bcmeireles, line 106: Are these wrong fixtures too or things Dremio does not support? If not supported I'd say we could just remove this instead of just commenting out

tests/functional/adapter/unit_testing/test_unit_testing.py
```diff
@@ -19,65 +19,206 @@
...
+        rows:
+          - {{ tested_column: {yaml_value} }}
+    expect:
+      rows:
+        - {{ tested_column: {yaml_value} }}
+"""
 
 class TestDremioUnitTestingTypes(BaseUnitTestingTypes):
+    # TODO: Revisit once we support unit testing with models that have dependencies
+    @pytest.fixture(scope="class")
```
bcmeireles, line 48: Let's create an internal ticket and have the DX as part of this `TODO` to keep track of this

bcmeireles (commented):
Good job ^^ don't forget the changelog and to mention the limitation of not supporting referencing

tests/functional/adapter/unit_testing/test_unit_testing.py
99Lys: There are some that Dremio support, others that Dremio doesn't support directly. For instance, Dremio doesn't support literal STRUCT, but there are other ways to achieve the same result, so I suppose we could leave them there and adjust later to the right format

dbt/include/dremio/macros/materializations/unit.sql
```diff
@@ -0,0 +1,42 @@
...
+
+  {%- set target_relation = this.incorporate(type=relation_type) -%}
+  {%- set temp_relation = make_temp_relation(target_relation) -%}
+  {%- set empty_sql = get_empty_subquery_sql(sql) -%}
+
+  {% do run_query(
+      get_create_view_as_sql(temp_relation, empty_sql)
+      if temp_relation.type == 'view'
+      else get_create_table_as_sql(True, temp_relation, empty_sql)
+  ) %}
```
99Lys, line 19: This implementation overrides the original one from `dbt-adapters` in order to enable unit testing on **views**. These were the lines added.

Unit testing feature apparently relies on the creation of a temporary copy of the relation being tested and stores it where the original relation is. In the original implementation, the relations being tested were always assumed to be **tables**, which led temporary copies to always be **tables** as well. Therefore, if a user wanted to test a view, it would attempt to create a **table** in a **space** (which is the "database" of a view).

So the fix consists of checking whether we have a **view** or a **table** by checking if the model's `database` corresponds to a space (`database`) or a source (`datalake`) and create the temporary copies accordingly.

valterfrancisco-dremio (approved):

simonpannek (approved):
Great work!

Issue #145, opened 2023-03-02
Replace ilikes in dbt-dremio to enhance performance


ArgusLi:
### Describe the enhancement requested

Investigate which ilikes can be replaced with equalities without leading to errors.
Replace ilikes with equalities.
Make equalities the default.

### Justification for this enhancement

Replacing the ilike in `dremio__list_relations_without_caching` with equalities has led to a significant performance enhancement https://github.com/dremio/dbt-dremio/issues/134. Are there any other ilikes that are hampering performance?

PR #292, opened 2025-04-22
Replace ilike with direct comparison

valterfrancisco-dremio:
### Summary

Replace ilike with direct equality where applicable

### Test Results

Manual tests and all automatic tests passing

### Related Issue

https://github.com/dremio/dbt-dremio/issues/145


dbt/include/dremio/macros/adapters/metadata.sql
```diff
@@ -212,7 +212,13 @@ limitations under the License.*/
   {% set sql %}
     select substring(schema_name, position('.' in schema_name) + 1)
     from information_schema.schemata
-    where ilike(schema_name, '{{ schema_name_like }}')
+
+    {%- if var('dremio:exact_search_enabled', default=false) %}
```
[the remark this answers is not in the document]
simonpannek: Yeah, this is an interesting one. Do you have a usage example where this would be triggered?

99Lys (commented):
Nice job!! No blockers on my side, but just wanted to confirm if we already have tests for this flag. It would be nice to include / update them

dbt/include/dremio/macros/adapters/metadata.sql
```diff
@@ -290,7 +304,7 @@ limitations under the License.*/
           end) as table_schema
           ,lower(table_type) as table_type
       from information_schema."tables"
-      {%- if var('dremio:exact_search_enabled', default=false) %}
+      {%- if var('dremio:exact_search_enabled', default=true) %}
         where table_schema = '{{ schema_name }}'
```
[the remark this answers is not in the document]
valterfrancisco-dremio: I manage to test the macro by calling them directly and nothing broke but I'm not confident enough, especially if the function is called by dbt internal functions because I've seen weird behaviour. Will change it to false.

valterfrancisco-dremio:
> Nice job!! No blockers on my side, but just wanted to confirm if we already have tests for this flag. It would be nice to include / update them

Add tests for the flag 👍

tests/functional/adapter/dremio_specific/test_exact_search_enabled.py
simonpannek, line 1: Hey! So this just tests whether settings the flag is working. We'd rather want to test whether `exact_search_enabled` being set will cause Dremio to do direct lookups.

99Lys (dismissed):
@valterfrancisco-dremio and I discussed about whether `dremio:exact_search_enabled` flag should be applied in all cases where `ilike` appears. There seems to be a case where `ilike` is being used but the flag doesn't exist in dbt environment yet. An internal ticket will be created to handle specific cases like this

tests/functional/adapter/dremio_specific/test_exact_search.py
99Lys, line 1: Based on what we've discussed about not being possible to grab the compiled sql of a macro that behind the scenes is performing direct comparison or ilike, I suppose your approach for this was to check if exact search is possible or not. 
For the reason I mentioned, I support you on this but you removed the flag test as well. I still feel it's important to keep the flag tests. 
Let me know what you think about this @valterfrancisco-dremio @simonpannek @bcmeireles

99Lys: Also please add a description of your changes in the changelog

simonpannek: This test doesn't have a dependency on the feature, so we're not testing any of the altered macros at all. I'm not sure what we're testing here exactly to be honest.

valterfrancisco-dremio: Updated the changelog and re added tests for exact search flag

simonpannek (approved):

99Lys (approved):
Nice job!!

PR #293, opened 2025-05-05
Ability to perform unit tests on models with dependencies - fixing ref() macro

99Lys:
### Summary

Fixing our custom `ref()` macro in order to support unit testing on models with dependencies.

### Description

This is originated from the error we were getting while trying to perform unit tests on models with dependencies → `Compilation Error ... 'dict object' has no attribute 'nodes'`. Our original override of `ref()` macro relies on `graph` Jinja function in some specific cases to retrieve the metadata of a model. 
However, this `graph` is not available (with content) during unit tests, but it's essential for other cases (e.g. **reflection** models). Therefore, we added a check before its usage and enabled `ref()` on unit testing models. With this fix, we basically are using the builtin `ref()`'s result.

### Test Results

- Refactored tests for unit testing
- All tests are passing

### Changelog


howareyouman (approved):
Great work with ref() function support!

tests/functional/adapter/unit_testing/test_unit_testing.py
```diff
@@ -106,119 +75,8 @@ def data_types(self):
...
-        return {
-            "my_model_2.sql": my_model_2_sql,
-            "unit_tests_2.yml": test_my_model_2_yml,
-        }
-
-    def test_case_insensitivity(self, project):
-        results = run_dbt(["run"])
-        assert len(results) == 1
-
-        results = run_dbt(["test"])
```
bcmeireles, line 173: why are we removing this test?

99Lys: This test was an override to the original `test_case_insensitivity` from dbt because at that time we didn't support `ref()` in unit testing so I needed to manually create a model without dependencies (I combined the two existing fixtures into one) and adjust the number of models being ran - `assert len(results) == 2` → `assert len(results) == 1`.
Since now `ref()` it's fixed, we can just run the original tests without adjustments

simonpannek (approved):

bcmeireles (approved):

valterfrancisco-dremio (approved):

PR #294, opened 2025-05-14
Fix data type issue in fixtures creation

99Lys:
### Summary

Fixes data type casting issues while mocking unit testing data.

### Description

We noticed that not all data types were being correctly parsed from dbt to Dremio in fixtures creation. This was happening because:
- dbt assumes all castings have the format `cast <data> as <type>` but for Dremio SQL dialect this won't always work (e.g. arrays) → therefore, we overrode `safe_cast` macro and adjusted the format of each casting according to the type / context of the query.
- The `input` field we provide in unit tests configuration will internally become an ephemeral model built based on the model being tested. This input model creation relies on the metadata stored in Dremio - column names, types, etc -, so in order to do this mapping from Dremio results to a table we can manipulate in dbt's environment, `agate` library was used. The problem is this library **infers** the types of the column values in a specific order by default ([agate.TypeTester](https://agate.readthedocs.io/en/latest/api/type_tester.html)) which led to the numeric type casting issues (values like _0_, _1_ and _null_ were all being treated as **Boolean**) → the fix is to force it to use a specific data type in types where the values that can be misinterpreted as Boolean.

### Test Results

- Refactored unit testing data types test → based on Dremio supported types, added tests for them
- All tests are passing

### Changelog


dbt/adapters/dremio/api/cursor.py
```diff
@@ -180,13 +180,16 @@ def _populate_results_table(self):
...
             for col in self._job_results["schema"]:
                 name = col["name"]
                 data_type_str = col["type"]["name"]
-                if data_type_str == "BIGINT":
-                    tester = agate.TypeTester(force={f"{name}": agate.Number()})
-
+                if data_type_str == "BIGINT" or data_type_str == "INTEGER":
+                    force[name] = agate.Number()
+            if force:
+                tester = agate.TypeTester(force=force)
```
bcmeireles, line 192: Have other adapters done such a thing? Trying to understand the logic behind `force`

99Lys: As far as I know other adapters didn't do anything like this (this might be specific to the way the adapter communicates with Dremio). In this case this "forcing" was already being done with `BIGINT` before, and I just added `INTEGER` as well

dbt/include/dremio/macros/utils/safe_cast.sql
```diff
@@ -0,0 +1,9 @@
+{% macro dremio__safe_cast(field, type) %}
+    {% if type.startswith('interval') or (field|string).startswith("'convert_from") %}
```
simonpannek, line 2: I've seen other connectors do these checks with `upper()`/`lower()` checks. Is this something we need to do as well, or are the types always lower case when working with Dremio (and if so, why)?

dbt/adapters/dremio/api/cursor.py
simonpannek: Could you explain in more detail what's happening here exactly?

dbt/include/dremio/macros/utils/safe_cast.sql
99Lys: Good point, we need. All the tests were in lowercase that's why is was working :')
Will fix it now

dbt/adapters/dremio/api/cursor.py
99Lys: Take a look at the description (or if you already did, let me know if there is anything that's not clear)

dbt/include/dremio/macros/utils/safe_cast.sql
```diff
@@ -0,0 +1,9 @@
+{% macro dremio__safe_cast(field, type) %}
+    {% if type.lower().startswith('interval') or (field|string).lower().startswith("'convert_from") %}
+        {{field}}
+    {% elif type == 'array' %}
+        {{type ~ field}}
```
howareyouman, line 5: How will this work? What is  `{{type ~ field}}`? Could you please add the example here?

howareyouman (approved):
Great result! 
I see only 1 type is missing - Map (I suppose because we can't have it in SQL without a table with this type). 
The description of PR was useful!

simonpannek (approved):

dbt/include/dremio/macros/utils/safe_cast.sql
99Lys: Suppose the query is
```sql
select array[1,2,3] as test_column
```
- `type` is `array`
- `field` is `[1,2,3]`
- `~` is used to concatenate strings

So the result will be `array[1,2,3]` (basically what is in the original query)

PR #296, opened 2025-05-28
dbt 1.9 migration

99Lys:
### Summary

Updating dbt-dremio to match dbt-core v1.9.0.

### Description

Following [Adapter Maintainers: Upgrading to dbt-core v1.9.0](https://github.com/dbt-labs/dbt-core/discussions/11125):
- Snapshot improvements
   - Snapshots: Enable setting a datetime value for `dbt_valid_to` for current records instead of `NULL`
   - Snapshots: Enable `hard_deletes` config to add a metadata column if a record has been deleted (automatically inherited from dbt-adapters)
   - Snapshots: Allow customizing names of metadata fields
   - Snapshots: Enable `unique_key` as a list (automatically inherited from dbt-adapters)
- Microbatch incremental strategy → optional, so not included in this PR

### Test Results

- Fixed _tests/hooks/test_run_hooks.py_
- Added `TestSnapshotDbtValidToCurrentDremio`, `TestSnapshotColumnNamesDremio`, `TestSnapshotColumnNamesFromDbtProjectDremio` and `TestSnapshotInvalidColumnNamesDremio` tests for the snapshot improvements

### Changelog

### DIFF
--- a/dbt/include/dremio/macros/adapters/snapshot.sql
+++ b/dbt/include/dremio/macros/adapters/snapshot.sql
@@ -17,26 +17,44 @@ limitations under the License.*/
   {%- set columns = config.get("snapshot_table_column_names") or get_snapshot_table_column_names() -%}
   {%- set insert_cols_csv = insert_cols | join(', ') -%}
 
+  {%- set update_source -%}
+    (
+      select *
+      from {{ source }}
+      where dbt_change_type in ('update', 'delete')
+    )
+  {%- endset -%}
+
+  {%- set insert_source -%}
+    (
+      select *
+      from {{ source }}
+      where dbt_change_type = 'insert'
+    )
+  {%- endset -%}
+
+  -- First, handle updates/deletes for matched records
   merge into {{ target }} as DBT_INTERNAL_DEST
-  using {{ source }} as DBT_INTERNAL_SOURCE
+  using {{ update_source }} as DBT_INTERNAL_SOURCE
+  on DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }} = DBT_INTERNAL_DEST.{{ columns.dbt_scd_id }}
+    {%- if config.get("dbt_valid_to_current") %}
+    and (
+      DBT_INTERNAL_DEST.{{ columns.dbt_valid_to }} = {{ config.get('dbt_valid_to_current') }}
+      or DBT_INTERNAL_DEST.{{ columns.dbt_valid_to }} is null
+    )
+    {%- else %}
+    and DBT_INTERNAL_DEST.{{ columns.dbt_valid_to }} is null
+    {%- endif %}
+
+  when matched then update
+    set {{ columns.dbt_valid_to }} = DBT_INTERNAL_SOURCE.{{ columns.dbt_valid_to }};
+
+  -- Then, handle inserts for non-matched records
+  merge into {{ target }} as DBT_INTERNAL_DEST
+  using {{ insert_source }} as DBT_INTERNAL_SOURCE
   on DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }} = DBT_INTERNAL_DEST.{{ columns.dbt_scd_id }}
 
-  when matched
-    {%- if config.get("dbt_valid_to_current") %}
-      and (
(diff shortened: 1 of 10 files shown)


tests/functional/adapter/empty/test_empty.py
```diff
@@ -41,15 +41,3 @@ def models(self):
             "model.sql": _models.model_sql,
             "sources.yml": schema_sources_yml,
         }
-
-    def test_run_with_empty(self, project):
```
99Lys, line 45: Removing this to keep the original `test_run_with_empty` test (basically the same, but without the comments)

dbt/include/dremio/macros/builtins/builtins.sql
```diff
@@ -24,8 +24,8 @@ limitations under the License.*/
       and model.config.format is defined
       else none -%}
     {%- set format_clause = format_clause_from_node(model.config) if format is not none else none -%}
-    {%- set relation2 = api.Relation.create(database=relation.database, schema=relation.schema, identifier=relation.identifier, format=format, format_clause=format_clause) -%}
-    {{ return (relation2) }}
+    {%- set relation2 = api.Relation.create(database=relation.database, schema=relation.schema, identifier=relation.identifier, format=format, format_clause=format_clause, limit=relation.limit) -%}
```
99Lys, line 27: Fix needed for _test_empty_: `--empty` flag applies `limit 0` to the relation but we were not applying it properly before

dbt/include/dremio/macros/adapters/snapshot.sql
```diff
@@ -14,22 +14,33 @@ limitations under the License.*/
...
-        then insert ({{ insert_cols_csv }})
-        values
-            ({% for column_name in insert_cols -%}
-                DBT_INTERNAL_SOURCE.{{ column_name }}
-                {%- if not loop.last %}, {%- endif %}
-            {%- endfor %})
-
-{% endmacro %}
\ No newline at end of file
+  {%- set columns = config.get("snapshot_table_column_names") or get_snapshot_table_column_names() -%}
```
99Lys, line 17: Snapshots: Allow customizing names of metadata fields → https://github.com/dbt-labs/dbt-adapters/pull/290/files#diff-f2707a5031a3e4b6d1455d3afa33315cc491e7dcde2b2ee07886bea29844ac10

howareyouman: Do we have a test that will validate this new behavior?

howareyouman (commented):
Could you please add the test for the new behavior of snapshot macro? 
The rest is great.

dbt/include/dremio/macros/adapters/snapshot.sql
howareyouman: But I don't see that we are using this BaseSnapshotColumnNames test from dbt-core here - https://github.com/dremio/dbt-dremio/blob/main/tests/functional/adapter/materialization/test_snapshots.py

howareyouman: My only concern here is to have a test that will test this macro with a new logic in it. 
If we already have tests which are covering it - I'm ok, but if not that we should create some.

tests/functional/adapter/materialization/test_snapshots.py
```diff
@@ -97,3 +103,15 @@ def project_config_update(self):
             "seeds": {"+twin_strategy": "prevent"},
             "name": "snapshot_strategy_timestamp",
         }
+
+class TestSnapshotDbtValidToCurrentDremio(BaseSnapshotDbtValidToCurrent):
```
howareyouman, line 107: Don't we need to override unique_schema and dbt_profile_data functions?

howareyouman: I mean should we keep them in TestSnapshotTimestampDremio and TestSnapshotCheckColsDremio classes as well?

howareyouman (dismissed):
Thanks for your PR!
It would be awesome if we could clean TestSnapshotCheckColsDremio and TestSnapshotTimestampDremio test classes.

howareyouman (approved):
I'm ok having duplication of tests from dbt-adapters, but at the same time, if it's possible to patch the problem there, we should do it.

valterfrancisco-dremio (approved):

